Issue #34, opened 2021-03-17
Update to use PySTAC


matthewhanson:
Currently messages are treated simply as dictionaries, with no STAC library (it had earlier used sat-stac before switching to straight JSON).

cirrus-lib should accept PySTAC entities where appropriate, and the core Lambdas at least should use PySTAC which will help with validation and make other things possible in the future, like maintaining a static STAC catalog.

matthewhanson:
Most of Cirrus uses straight dictionaries of STAC Items. This is limiting in that we aren't taking advantage of the validation and convenience features of PySTAC. However, it may be that using dictionaries is faster, and sufficient.

Review code in cirrus and in cirrus-lib to identify where using PySTAC vs dictionaries may be a benefit.

jkeifer:
@matthewhanson Is this still issue still applicable?

matthewhanson:
No, stac-task uses PySTAC, and you get input items as PySTAC Items or as dictionaries

PR #41, opened 2021-05-09
[Cirrus-37] Replaced sat-search with pystac-client

quetcodesfire:
I removed the sat-search requirement and replaced it with pystac-client. In addition to this, I also added pystac-client to all of the core lambdas as I see there are plans to implement it in issue #34.


matthewhanson:
The code in the feeders/stac-api/feeder.py file also needs to be updated to use pystac-client.  It works more or less the same, but the import needs to be changed and the function names are slightly different.

Closing for now, can open up new PR

PR #44, opened 2021-05-19
[Cirrus 37] Replaced Sat-Search Code with Pystac_Client

quetcodesfire:
I previously brought pystac_client into the project, in this PR I have removed sat-search and replaced that code with pystac_client.


matthewhanson:
With the recent refactor of Cirrus, this PR would need to be updated.

We also need to rethink how this should be implemented, given that this is a very common way to use Cirrus, so perhaps it should be included in the Cirrus `process` function as a valid payload.

Also, request splitting and async http calls will be part of the next major release of pystac-client which will make a lot of the stac-api feeder code obsolete.

We'll revisit this code with the next version of pystac-client.

PR #49, opened 2021-10-03
cirrus cli

jkeifer:
Here it is, at long last.

It still needs some testing with the large amount of changes I've made in the past 24 hours, but it is mostly here and decent.

Some outstanding items worth mentioning:
* The cloudformation parsing is ill-informed and not great. Resources and Outputs work, but if you need any other cloudformation stuff it won't work. I have some ideas around how to fix this, but it is probably a medium-large lift.
* The tack batch resources still have not been ported over to the task definitions (waiting to do this until we have a release we can target for dependent projects to keep them unchanged until ready).
* The github workflow testing isn't running on my branch, so I don't know if that workflow config is correct.
* I want to try to merge cirrus-lib into this once again. Reading the docs a couple more times, I think I can force the serverless python requirements to do my bidding and only include the lib and its deps in lambda packages, not any cli deps. But it'll take some testing to verify.

I am also sure changelog/readme/whatever docs need serious help after these changes. I am not sure how best to approach them; perhaps a team effort would be best.

Lastly, calling this a 1.0 release, even if alpha, seems generous. I would feel much more comfortable with a 0.x release; aren't minor versions of pre-1.0 releases not guaranteed to be compatible, so we don't have to go all the way to 1.0, right?


matthewhanson:
This is great @jkeifer. It's such a large change I think it's best to get it merged into main and test from there, making fixes with subsequent PRs rather than trying to continue working from this branch.

> * The cloudformation parsing is ill-informed and not great. Resources and Outputs work, but if you need any other cloudformation stuff it won't work. I have some ideas around how to fix this, but it is probably a medium-large lift.
> * The tack batch resources still have not been ported over to the task definitions (waiting to do this until we have a release we can target for dependent projects to keep them unchanged until ready).

Can you create tickets for these in this repo?

> * The github workflow testing isn't running on my branch, so I don't know if that workflow config is correct.

There's no github workflows set up. We can set one up for testing if we can get tests working offline.  We can add a workflow for PyPI publishing on release, I created a ticket.

> * I want to try to merge cirrus-lib into this once again. Reading the docs a couple more times, I think I can force the serverless python requirements to do my bidding and only include the lib and its deps in lambda packages, not any cli deps. But it'll take some testing to verify.

So do you mean cirrus-lib will just be included in this new version of cirrus?

> I am also sure changelog/readme/whatever docs need serious help after these changes. I am not sure how best to approach them; perhaps a team effort would be best.

Add some line items to the CHANGELOG describing the change.  No need to get real specific at this point until after we release this version and have docs.

> Lastly, calling this a 1.0 release, even if alpha, seems generous. I would feel much more comfortable with a 0.x release; aren't minor versions of pre-1.0 releases not guaranteed to be compatible, so we don't have to go all the way to 1.0, right?

Semantic Versioning:
***
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards compatible manner, and
PATCH version when you make backwards compatible bug fixes.
Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.
***

I think it's fine calling this a MINOR version, as it's just creating a CF template the same as before and can be used to update existing installations, so it would be backward compatible.

So we'll shoot for a 0.5.0 release, there's some tickets to create, and we can discuss what version to target. I'm going to go through the tickets now and mark any relevant ones for 0.5.0

matthewhanson (approved):

PR #60, opened 2021-10-04
#50: python-publish gh workflow

jkeifer:
We'll need to make sure the secrets are configured for this to work, but here is an initial pypi publish workflow. The version used for the release is set by the tag name, so simply name a release with a version like "v1.2.3" and that'll be the version used in the package.

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -1,0 +1,29 @@
+name: Upload Python Package
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  deploy:
+
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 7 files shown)


trevorskaggs (approved):
:shipit:

jkeifer:
Sorry to keep adding on to this PR, but I've been fixing up a number of things I think we'll want to ensure are in place before merging this PR. That said, we should also get #63 merged first, as we'll want that to go out with any release we make.

jkeifer:
#63 has been merged, so I think we just need this PR (and pypi secrets) to get a v0.5.0a0 release out.

matthewhanson (approved):

Issue #59, opened 2021-10-04
Lambda as batch fails to find module


jkeifer:
Running lambdas as batch results in the error `ModuleNotFoundError: No module named 'lambda_function'`. It appears renaming the `task.py` and `feeder.py` files to `handler.py` broke the import in the batch image.

I was unaware that the AWS convention is `lambda_function.lambda_handler` when I chose `handler.handler`. Given this issue and that naming convention, I think it would be best to rename everything to the former.

PR #63, opened 2021-10-04
standardize on lambda_function.lambda_handler #59

jkeifer:
Resolves the issue in #59.

### DIFF
--- a/cirrus/cli/components/base/_lambda.py
+++ b/cirrus/cli/components/base/_lambda.py
@@ -38,7 +38,7 @@ class Lambda(Component):
         if not hasattr(self.lambda_config, 'module'):
             self.lambda_config.module = f'lambdas/{self.name}'
         if not hasattr(self.lambda_config, 'handler'):
-            self.lambda_config.handler = 'handler.handler'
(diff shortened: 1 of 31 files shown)


matthewhanson:
I'm good with merging this so go ahead. Keeping to the default lambda handler naming convention is one less boilerplate item we have to manage.  

However, be aware that the cirruslib.logger module allows control of logging based on module names, so this breaks that feature.

https://github.com/cirrus-geo/cirrus-lib/blob/main/cirruslib/logging.py#L26

However, in practice all these loggers tend to have the same configuration. I think logging in general should be reconsidered, with the main focus to be the ability to track a single job through all of a step functions tasks via filterable JSON log messages.

matthewhanson (approved):

jkeifer:
I looked at the logging configuration and I am insuring the logger name is setup as intended in the `lambda_function.py` when creating a new task/feeder/function, and all existing lambdas have been manually updated to have the desired logger name hardcoded.

That said, my thoughts around class-based handlers would solve the logger thing in a more dynamic, less-boilerplate way. So I think this could get much easier to manage in the future.

trevorskaggs (commented):
:shipit:

PR #76, opened 2021-11-08
74/workflow chaining

jkeifer:
This is a sketch of the changes to implement workflow chaining. I based this branch on the branch of #69, which merges in cirrus-lib, so that I could make all the required changes in one place. If it turns out we are not good with #69 I can port the cirrus-lib changes into the cirrus-lib repo.

I plan to work on some tests to validate the behavior will work as intended, but essentially it should allow:
* a process block array, so we can chain workflows together, each with a specific process definition
* a nested array within that to allow parallel workflow fan-out, where workflow outputs branch

The intent of all this is to allow a process block that looks roughly like this:
```
...
    process = {[
        { workflow 1 },
        [{ workflow 2a }, { workflow 2b }],
        { workflow 3 }
    ]}
...
```

This would run like :
```
    workflow 1  --+-->  workflow 2a  ---->  workflow 3
                  |
                  +-->  workflow 2b  ---->  workflow 3
```

Currently the fan-out happens in the `update-state` step after a workflow has finished, though I believe I should move it into the `process` lambda handling of input payloads, to support fan-outs for the first workflow process definition.

A future improvement will be designing some mechanism to wait for parallel outputs required for a later workflow step, which would allow us to model a chain like:
```
    workflow 1  --+-->  workflow 2a  ---->  workflow 3  --+
                  |                                       +-->  workflow 4
                  +-->  workflow 2b  ---->  workflow 3  --+
```

But that's out of scope at the moment.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ pyyaml~=5.4
 click~=8.0
 rich~=10.6
 cfn-flip~=1.2
-cirrus-lib>=0.6.0a1
+cirrus-lib>=0.6.0a2
--- a/src/cirrus/builtins/functions/process/lambda_function.py
+++ b/src/cirrus/builtins/functions/process/lambda_function.py
@@ -53,6 +53,6 @@ def lambda_handler(payload, context):
 
     if len(catalogs) > 0:
         cats = Catalogs(catalogs)
-        cats.process()
+        cats()
 
     return len(catalogs)
--- a/src/cirrus/builtins/functions/update-state/lambda_function.py
+++ b/src/cirrus/builtins/functions/update-state/lambda_function.py
@@ -12,6 +12,7 @@ from cirrus.lib.logging import get_task_logger
 logger = get_task_logger('lambda_function.update-state', catalog=tuple())
 
 # envvars
+PROCESS_SNS_TOPIC = getenv('CIRRUS_QUEUE_TOPIC_ARN', None)
 FAILED_TOPIC_ARN = getenv('CIRRUS_FAILED_TOPIC_ARN', None)
 INVALID_TOPIC_ARN = getenv('CIRRUS_INVALID_TOPIC_ARN', None)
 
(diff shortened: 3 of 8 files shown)


jkeifer:
After further reflection I have decided that the fan-out support for parallel workflows is misguided, at least at this point in time, given the lack of a solid use-case. Of particular note, such parallelization would not provide a mechanism to modify the items submitted to the workflow, such that all parallel invocations would get the same set of items. As it stands, this is incompatible with the normal state DB check to see if a given set of items is currently processing or has already been completed.

jkeifer:
I'll be splitting the cirrus-lib changes out of this branch and making them in cirrus-lib, as we have decided not to merge #69 and will instead continue maintaining cirrus-lib separately.

jkeifer:
> After further reflection I have decided that the fan-out support for parallel workflows is misguided, at least at this point in time, given the lack of a solid use-case. Of particular note, such parallelization would not provide a mechanism to modify the items submitted to the workflow, such that all parallel invocations would get the same set of items. As it stands, this is incompatible with the normal state DB check to see if a given set of items is currently processing or has already been completed.

Well, we have an upcoming use case for this, and I have to say that I was considering this a bit wrongly. It will work, as long as the parallel fan out is for multiple unique workflows. For example, when using a lower level of data as input to multiple next-level workflows, this feature would be rather useful.

This said, I don't think doing the fan-out in `process`, as I previously proposed, would actually be better than in `update-state`. I think if `process` receives a payload, it shouldn't modify that payload, else testing just one branch of the fan-out becomes problematic, when required. It is also easier to put the onus on the user to queue multiple payloads for each desired branch when doing so at the start of the chain, whereas workflow chaining requires some mechanism to enable the fan-out, with `update-state` effectively doing the "user" task of splitting the fan-out.

jkeifer:
Once the cirrus-lib side https://github.com/cirrus-geo/cirrus-lib/pull/32 is merged we'll need to cut a release and update the requirements.txt with that new version, before merging this PR.

trevorskaggs (approved):
Dependent on merging of https://github.com/cirrus-geo/cirrus-lib/pull/32, but :shipit:

Issue #75, opened 2021-11-04
post-batch does not log its own errors


jkeifer:
I saw a situation where post-batch failed to pull the error from a batch job. However, I was unable to troubleshoot, as post-batch also fails to log its own errors. That needs to be fixed so we can address any other issues.

PR #86, opened 2021-12-14
post-batch cleanup and fixes re 75

jkeifer:
* Ensures exceptions within post-batch are logged
* No longer drops chars at beginning of error messages:
  * Uses `removeprefix` as intended, instead of `lstrip`
* More robust error class determination:
  * Now matches on whole word at start of message text ending with colon
  * Still susceptiable to messages like `Error:.*`
* Other cleanup

### DIFF
--- a/cirrus/cli/components/tasks/config/post-batch/lambda_function.py
+++ b/cirrus/cli/components/tasks/config/post-batch/lambda_function.py
@@ -4,10 +4,15 @@ import json
 import boto3
 
 from cirruslib import Catalog
+from cirruslib.logging import get_task_logger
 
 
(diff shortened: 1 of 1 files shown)


jkeifer:
Made this a draft as I haven't tested the lambda code yet.

jkeifer:
Tested these changes and made some fixes. Note that actuallly logging any exceptions in the exception handler for #75 will require the cirrus-geo/cirrus-lib#30 change, but this PR can be merged without pulling in a new cirrus-lib version. We'll just want to make sure we have a new cirrus-lib release and we're using it before cutting the next release in this project.

trevorskaggs (approved):
Looks like a good change! Good find on the `lstrip` modification.

Issue #54, opened 2021-10-04
Rename Cirrus CLI Collections?


matthewhanson:
While the name `Collections` certainly make sense to describe sets of deployable resources, it unfortunately collides with with Collections as used in STAC. Since Cirrus is fundamentally a STAC-based infrastructure and the term Collections is widely used to describe collections of items published by Cirrus, I think this could lead to some understandable confusion.

Not sure of the best action to take here...if this is really just an internally used class that the vast majority of users need not know then it may be OK. Otherwise it may be prudent to at least rename the `Collections` class to something more descriptive.

Issue #53, opened 2021-10-04
Better separation of cli/deploy code from tasks/feeders


matthewhanson:
I'm not fond of the all the modules falling under a cirrus/cli package, since there's a difference between task code that may be used by the CLI (which may be used without the CLI) and code that is the actual CLI.  I'd like to see a better separation of packages here. 

For instance, we roughly have something like:
- cirrus-deploy: code that uses other code and templates to create a CF template
- cirrus-cli: the actual CLI code that calls cirrus-deploy functions
- cirrus-components: this is already a logical grouping, except it's under cli when they aren't really CLI specific.

These are just initial thoughts, we can talk through.

trevorskaggs:
👍  to `cirrus-components` recommendation.

PR #87, opened 2021-12-15
53/reorg

jkeifer:
This PR separates out the builtin components/resources into their own package and splits the cli and core code into separate packages as well, closing #53.

The separation of the cli and core code is not totally complete; attempting this reorg makes it obvious where core and cli concerns have been conflated and shows we have some work to do to truly create a standalone core package free of cli baggage. That work is not immediately necessary, however, so I am punting on it for now. We can reevaluate in the future if we deem further work here necessary (if it doesn't otherwise happen as part of other future changes).

I also moved the src code into `src/`, as that seems to be a more preferred project structure these days.


jkeifer:
The diff is pretty rough. I'd recommend checking out the branch and reviewing the larger organizational pattern in that context.

jkeifer:
I've also included the renaming of the `collections` stuff to `groups` to close #54. I'm not super happy with `groups` as a name, but it seems serviceable. A thesaurus search didn't reveal any more appropriate choices, so I went with it (all other good ones already have specific connotations as well). I liked `gaggle`, but I figured that probably wouldn't have passed review. :)

trevorskaggs (approved):
Good set of changes, this will make future development more clear!

Issue #55, opened 2021-10-04
cirrus-lib dependency


matthewhanson:
The `requirements.txt` includes cirrus-lib, however it's not clear if cirrus-lib is really used by the Cirrus CLI, or if it is a dependency of tasks that are packages by the Cirrus CLI.

When packaging tasks does it include this requirements.txt file?  Is there a way to specify global requirements for all tasks?

jkeifer:
Cirrus-lib is a packaging dependency, and is currently not used by the cirrus CLI.

When setting up the lambda directories for packaging, a `requirements.txt` file created for each lambda. The contents of that file include any python dependencies listed under the `python_requirements` key in the lambda's `definition.yml` file, along with cirrus-lib. The version specified for cirrus-lib is extracted from the cirrus package metadata, such that the version in the cirrus `requirements.txt` will also be used for lambda packaging as a "global requirement".

Beyond that, we do not currently support global requirements for all lambdas (I say lambda rather than task here as all lambda-type components are packaged the same way at the moment).

trevorskaggs:
Potentially OBE by inclusion of cirrus-lib

jkeifer:
See #66 for cirrus-lib inclusion.

jkeifer:
#89 changes how the cirrus-lib dependency is handled. Now, we install cirrus-lib in all lambdas as part of the `cirrus build` process by copying the `cirrus.lib` python package into the cirrus build dir, and symlinking that into each of the lambda directories within. When serverless packages the lambdas, it bundles cirrus-lib in as well.

Some changes to allow this to work mean that the lambda `requirements.txt` files for the serverless python plugin no longer gets the cirrus-lib dependency version injected, rather all dependencies of cirrus-lib (based on package inspection using `importlib.metadata`) are injected, and cirrus-lib itself is omitted as it does not need to be pip-installed.

Part of making this all work, and looking further at the serverless python plugin made me realize that we already had a top-level `pythonRequirements` config in `cirrus.yml`, which per the plugin supports an `include` key, to provide a list of global dependencies for all lambda. I added support in the dependency injection (should I call it that despite its other connotations?) for such globally-defined requirements, even though using that feature appears to be an anti-pattern.

Anyway, cirrus-lib is still not used directly by cirrus-geo other than as the copy source for the lambdas. In that sense it is actually required to be installed to the python env now, whereas before it was not really required.

@matthewhanson I am going to close this issue, but if you have further concerns please reopen.

PR #89, opened 2021-12-22
build tests store less state and cirrus-lib from python env

### LINKED ISSUE #66: Merge cirrus-lib
As mentioned in #49, given the tight coupling of cirrus-geo and cirrus-lib and the fact that the former is now a python lib, we could merge the two together.

We need to be careful, however, that cirrus-geo does not become a dependency of the lambda build process, at least not in a way that would pull in cli dependencies. In other words, lambda deploy package size is a concern, and we should ensure we are not cluttering it unnecessarily.

jkeifer:
This PR encapsulates two sets of changes:
* rather than storing every file put into the build, we now just track changes to serverless.yml with a diff, and all other files simply by added/changed/removed (using shasum/size for change detection)
* rather than using python-serverless-requirements to pull cirrus-lib from pypi, we instead copy cirrus-lib from the local python env, and build that into the lambda packages.

The testing changes help with the cirrus-lib change by reducing the fixture diff (though all the fixture deletions are still part of this review, sorry). The cirrus-lib change closes #66 by _not_ merging cirrus lib but using the lessons learned from an attempt at doing so, and addresses #55 by better addressing how cirrus-lib is packaged and `pythonRequirements` are managed.

Note that in lambda `definition.yml` files, the `python_requirements` keyword has been renamed to `pythonRequirements` for consistency, and also moved under the `lambda` section. This is a non-backwards compatible change.

Lastly, note that this PR is dependent on the cirrus-lib changes in cirrus-geo/cirrus-lib#31. Until those changes are merged and a release is cut on that side, the CI tests will fail. At the time of a release there we will want to update the requirements.txt for this project as well, to ensure we are using that new release as the minimal version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,7 @@ node_modules/
 .serverless
 .dist/
 dist/
-build/
+/build/
 .env
 packages/
 *.egg-info/
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ pyyaml~=5.4
 click~=8.0
 rich~=10.6
 cfn-flip~=1.2
-cirrus-lib>=0.5.1
+cirrus-lib>=0.6.0a1
--- a/src/cirrus/builtins/feeders/feed-rerun/lambda_function.py
+++ b/src/cirrus/builtins/feeders/feed-rerun/lambda_function.py
@@ -4,8 +4,8 @@ import json
 import logging
 import sys
 
-from cirruslib.statedb import StateDB
-from cirruslib.utils import submit_batch_job
+from cirrus.lib.statedb import StateDB
+from cirrus.lib.utils import submit_batch_job
 from os import getenv
 
 
--- a/src/cirrus/builtins/feeders/feed-s3-inventory/lambda_function.py
+++ b/src/cirrus/builtins/feeders/feed-s3-inventory/lambda_function.py
@@ -12,7 +12,7 @@ from os import getenv, path as op
 
 import pyorc
 from boto3utils import s3
-from cirruslib.utils import submit_batch_job
+from cirrus.lib.utils import submit_batch_job
 
 
 # envvars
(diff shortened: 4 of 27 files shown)


cirrus/cli/project.py
```diff
@@ -126,15 +124,30 @@ def build(self) -> None:
...
         # write serverless config
-        self.config.build(self.collections).to_file(bd.joinpath(DEFAULT_SERVERLESS_FILENAME))
+        self.config.build(self.collections).to_file(
+            bd.joinpath(DEFAULT_SERVERLESS_FILENAME),
+        )
+
+        # copy cirrus-lib to build dir for packaging
+        lib_dir = bd.joinpath('cirrus', 'lib')
+        shutil.copytree(
+            lib.__path__[0],
```
jkeifer, line 142: change import to be cirrus.lib, as just lib is confusing

cirrus/cli/project.py
```diff
@@ -126,15 +124,30 @@ def build(self) -> None:
...
+            lib_dir,
+            ignore=shutil.ignore_patterns('*.pyc', '__pycache__'),
+            dirs_exist_ok=True,
+        )
+        try:
```
jkeifer, line 147: why is this symlink happening?

cirrus/cli/project.py
```diff
@@ -148,10 +161,14 @@ def build(self) -> None:
...
                 fn_dirs.add(outdir)
-                fn.link_to_outdir(outdir, cirrus_req)
+                fn.link_to_outdir(outdir)
+                outdir.joinpath('cirrus').symlink_to(
+                    misc.relative_to(outdir, lib_dir.parent),
+                )
 
         # clean up existing but no longer used lambda dirs
         for d in existing_dirs - fn_dirs:
+            print(d)
```
jkeifer, line 171: remove print

trevorskaggs (dismissed):
:shipit:

trevorskaggs (approved):
Thanks for the walk through, good set of changes, 🚢 🇮🇹

PR #90, opened 2021-12-22
rework build test to not have copies of all files

jkeifer:
When testing the files generated by the build step, we really only care to diff the content of the serverless.yml file. For all other files we can simply maintain a list of the shasums and sizes, and any differences will tell us if a file has been added, removed, or changed.

The advantage here is mainly much simpler code reviews and less redundant files maintained in the git repo. It comes at the expense of not being able to see a diff of changes outside the serverless.yml file, so questions about actual file content will need to be answered via manual inspection in a separate test project. While that disadvantage is significant, I believe the reduction of state stored in git per the stated advantage still outweighs it. Particularly with #89, which pulls in cirrus-lib and causes even more build files.


trevorskaggs (approved):
Thanks for splitting this out! :shipit:

PR #91, opened 2021-12-30
Better cloudformation support re #64

### LINKED ISSUE #64: Better cloudformation support
We currently support cloudformation for resource and output templates, but the division between resources and outputs in that way is rather forced, and moreover, we don't support the full spectrum of cloudformation top-level keys (see https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-anatomy.html).

In short, the cloudformation handling could use some reimagining. I think continuing to manage it at the "object" level ("a resource", "an output", etc) is desired, but beyond that I'm not sure we have any particular constraints.

jkeifer:
Now we support all cloudformation top-level keys (except template version and template description) for both cloudformation .yml files and cf defined in batch tasks.

Some key notes:
* cloudformation yml files are now consolidated under `/cloudformation/`, rather than `/resources/` and `/outputs/`
* the batch task definition syntax has changed to nest `Resources` under `resources`, the same as in serverless.yml,
to allow other cf types beyond just `Resources`

This PR also includes a number of small bugfixes and other improvements discovered while working on this feature. It is likely best to review the changes in context with the commit messages to help resolve any uncertainties.

Closes #64.


src/cirrus/core/components/tasks.py
```diff
@@ -17,23 +17,20 @@ def load_config(self):
...
             self.handler.validate(required=True)
 
-        # global env vars from cirrus.yml inherited in config build step
-        self.batch_env = self.config.get('environment', {})
         batch = self.config.get('batch', {})
         self.batch_enabled = batch.get('enabled', True) and self._enabled and bool(batch)
-        self.batch_resources = [
-            self.create_batch_resource(name, definition)
-            for name, definition in batch.get('Resources', {}).items()
+        self.batch_cloudformation = [
```
jkeifer, line 22: this needs to be hierarchical based on type

trevorskaggs (approved):
:shipit:

Issue #47, opened 2021-06-08
Make data bucket user configurable


matthewhanson:
The main Cirrus data bucket is where all the STAC Items and the data themselves will get uploaded to (unless a workflow specifies another location).

Right now it is named automatically with the normal serverless method - `cirrus-<stage>-<randomchars>`

Since users will often interface directly with the bucket, it seems reasonable that you'd want to have something more readable and specify the bucket name exactly.

- change default bucket name to `cirrus-<stage>-data`
- Update install/deploy docs to clearly indicate how to change the name
- Users may want to use an already existing bucket, also allow for specifying existing bucket and not creating a new one

trevorskaggs:
This is currently already possible by overriding in your projects cirrus.yml.  Need additional documentation to explain this process.

jkeifer:
I've looked into this issue a bit more and I have some findings.

1. If one overrides the `Data` bucket in a project with a different name it'll work as long as a bucket with that name does not already exist.
2. Changing the name of the `Data` bucket after the stack is initially created will result in a cloudformation error _if the `Data` bucket is not empty_. If it is empty, a new bucket will be created and the old bucket will be deleted. But if it is not empty, the new bucket will be created and the old bucket will not be able to be removed. However, I believe the new bucket will not be removed in the cloudformation rollback, so one can always copy the data to the new bucket and manually remove the old bucket. More testing would be required to confirm these details.
3. If overriding the `Data` bucket to specify a bucket that _already exists_, cloudformation will fail with an error because it cannot simply "adopt" existing resources. There appears to be a way to import existing resources (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-import-existing-stack.html), but I am struggling to see how it could work in this case. I believe we'd need a way to remove `Data` from the stack so that a new `Data` could be defined in the stack for that new resource. It would also appear to be a rather manual process of modifying an already-applied cloudformation template and ensuring those same changes are persisted in the cirrus project.

All this said, which bucket cirrus actually uses for data is controlled by an env var (typically set at the global level in cirrus.yml):

```
CIRRUS_DATA_BUCKET: !Ref Data
```

Setting this var to reference something other than `Data` will allow the user to use a different bucket without any weird cloudformation overrides/imports/etc. As long as users can accept having the `Data` bucket automatically created and existing in their account, then they can use whatever bucket they want as easily as changing this env var.

[...]

PR #92, opened 2021-12-31
Renaming and batch task conversion

### LINKED ISSUE #65: Conversion of batch tasks to use new per-task paradigm
We currently reuse batch resources across multiple tasks, as makes sense. However, with the addition of per-task batch resources in the task `definition.yml`, we need to convert all existing task that require batch support to have their own unique resources declared in their defintions.

### LINKED ISSUE #79: Renaming components
Rename 
- cirrus-lib Catalog to ProcessPayload
- from_payload to from_event
- SNS queue to process
-

jkeifer:
This PR tackles several issues:
* Renaming per #79 (requires a new cirrus-lib release/version once cirrus-geo/cirrus-lib#34 is merged)
* Renaming default S3 buckets per #47 
* Converting built-in tasks using batch to new batch task paradigm per #65
* Attempts to set minimal permissions, explicitly declaring them on lambdas where additional permissions are required
* Various config clean up/fixes
* Per-task batch jobs do not get specific env vars, but inherit from global/task env vars

Merging this should close #65 and close #79. #47 might need additional documentation changes, but we could consider closing that and expecting the documentation updates to be part of future work, at least with regard to the 0.5.0 release.

I am opening this as a draft as it needs significant testing in a real environment to ensure the new resources, permissions, env var changes, and lib changes all work as expected.


jkeifer:
It is expected the tests are currently failing, as I have not yet updated the build fixtures to reflect these changes. I figured it would be best to do all manual testing and make any changes before updating the fixtures.

jkeifer:
Testing revealed a number of bugs that need to be fixed up, but I eventually worked them all out and I think this is now in a good place. I also got the cirrus-lib 0.6.0 release out so I updated the requirements and the build test fixtures to get the tests passing.

trevorskaggs (approved):
Looking good, thanks for the updates!

PR #100, opened 2022-01-11
fix showing yaml fn args re #98

### LINKED ISSUE #98: Displaying yaml with functions fails to show arguments
For example, when running `cirrus show workflows cog-archive definition`, I see things like this:

```
    ...
    publish:
      Type: Task
      Resource:
        Fn::GetAtt:
      End: True
    ...
```

The display of those functions needs to be fixed.

jkeifer:
Ensures the content is escaped before printing to the terminal to preserve function args in `[]`. See https://rich.readthedocs.io/en/stable/markup.html#escaping for the rich documentation on the subject. Closes #98.

### DIFF
--- a/src/cirrus/core/components/files/base.py
+++ b/src/cirrus/core/components/files/base.py
@@ -86,4 +86,4 @@ class ComponentFile:
         path.write_text(self.content_fn(parent_component))
 
     def show(self):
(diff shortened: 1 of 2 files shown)


trevorskaggs (approved):
Send it! 🚢 📥

PR #101, opened 2022-01-12
Chaining item filter support

### LINKED ISSUE #99: Filtering items when chaining workflows
It seems like a reasonable expectation that a chained workflow might not want or need all the items produced by the initial workflow. We should add a filter functionality to the next workflow payload construction that will, for each next workflow, use some parameter in the process definition to filter the list of items before returning that new payload.

For example, a payload like this (syntax simplified for illustration purposes):

```
{
  items: [ a, b, c ],
  process: [
    { wf_1 },
    { wf_2: filter: "a*" }
  ]
}
```

may output something like:

```
{
  items: [ a1, a2, a3, b1, b2, c1, c2, c3 ],
  process: [
    { wf_1 },
    { wf_2: filter: "a*" }
  ]
}
```

and per the filter we would end up with a payload for the second workflow like:

```
{
  items: [ a1, a2, a3 ],
  process: [
    { wf_2: filter: "a*" }
  ]
}
```

jkeifer:
This PR pulls in a cirrus-lib version that supports item filtering when chaining workflows, dependent on the merge of cirrus-geo/cirrus-lib#37 and the v0.6.1 release. Until those two conditions are met the tests will fail.

Closes #99.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ and then replace all custom `cirrus.yml` content that is still required.
 
 - README badges and codecov support ([#88])
 - workflow chaining support ([#74])
+- item filter support for workflow chaining ([#99])
 - CloudFormation support for all types beyond Resources and Outputs ([#64])
 - `AWS_REGION` and `AWS_DEFAULT_REGION` env vars injected into batch job
   definitions by default ([44bebc5])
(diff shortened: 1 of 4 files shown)


jkeifer:
With the cirrus-lib 0.6.1 release and an update to the test fixture, this branch is ready to for review. Of course, the doc-update branch should be merged first.

trevorskaggs (approved):
Annnnd boom goes the dynamite. 💥

PR #112, opened 2022-02-08
output_payload in update-state functions

### LINKED ISSUE #111: Workflow chaining broken: uses step function input payload rather than output
In testing workflow chaining more fully, we found that the payload used in `update-state` is the step function's input payload, rather than the output payload. That made sense before chaining, as in an error case the output payload can be missing. However, now that we want to chain the output of the step function to the input of the next workflow, not using the output for that is obviously flawed.

jkeifer:


jkeifer:
Closes #111.

src/cirrus/builtins/functions/update-state/lambda_function.py
```diff
@@ -167,20 +166,22 @@ def parse_event(event):
             if status == FAILED:
                 error = get_execution_error(event['detail']['executionArn'])
             return (
-                ProcessPayload.from_event(json.loads(event['detail']['input'])),
+                ProcessPayload(json.loads(event['detail']['input'])),
+                ProcessPayload(json.loads(event['detail']['output']))
```
grmpflh27, line 170: since the `output_payload` can be `None`, while the `status == "SUCCEEDED"` you are opening yourself up for a bug` in `workflow_completed` (something like `'NoneType' object has no attribute 'next_payloads'`)

--> check for `None` and return early

src/cirrus/builtins/functions/update-state/lambda_function.py
```diff
@@ -154,9 +152,10 @@ def parse_event(event):
     #   - error object
     try:
         if 'error' in event:
-            logger.debug('looks like a stac record with an error message, i.e., workflow-failed')
+            logger.debug('looks like a payload with an error message, i.e., workflow-failed')
             return (
                 ProcessPayload.from_event(event),
+                None,
```
grmpflh27, line 158: I recommend to documenting/ updating the return types

src/cirrus/builtins/functions/update-state/lambda_function.py
```diff
@@ -191,4 +192,4 @@ def lambda_handler(event, context={}):
         logger.info("Status does not support updates")
         return
 
-    status_update_map[status](payload, error)
+    status_update_map[status](input_payload, output_payload, error)
```
grmpflh27, line 195: in order to **not** force the function signature between different state handlers (failed, completed) to be identical, I'd remove this handler map and just go with 

pseudo code:
```python
if status == FAILED:
    workflow_failed(item_id=input_payload['id'], error=error)  # NOTE: only pass what's needed of a complex obj (id)
    
if status == SUCCEEDED:
     workflow_succeeded(item_id=input_payload['id'], output_payload=output_payload)  # NOTE: only pass what's needed of a complex obj (id)

src/cirrus/builtins/functions/update-state/lambda_function.py
jkeifer: I thought long and hard about the value of output, and I couldn't see any circumstance where it would be null from a successful workflow execution. I was originally considering a more complicated way of handling that in the `parse_event` function, but decided the complexity was not worth it. That said, with this simpler way of returning both input and output payloads, you're right, it is easy enough to check if it has a value before using it in `workflow_completed`. I've added that check.

`workflow_completed` is so named because the corresponding cirrus state is `COMPLETED`, however I see how that is ambiguous with regard to this function name. I plan a fair bit of refactoring for this function in the near future, at which time I will either rename the function or replace it with entirely different semantics.

src/cirrus/builtins/functions/update-state/lambda_function.py
jkeifer: Agreed, but again I plan future refactoring which will make this moot.

trevorskaggs (approved):

PR #118, opened 2022-02-16
Add workflow aborted state handling

jkeifer:
Also handles `TIMED_OUT` step function events as a `FAILURE`.

Note that this PR is dependent on cirrus-geo/cirrus-lib#45 and the 0.7.0 release that will be made once that is approved and merged. The tests will not pass until that release is available.


trevorskaggs (approved):
:shipit:

PR #119, opened 2022-02-17
rmtree for lambda dirs on rebuild

### LINKED ISSUE #105: Lambda clean_outdir fails when rebuilding project from pypi package
It appears that directories in lambdas are not handled correctly on rebuilds. Initially dirs were symlinked into the lambda build dir, but that appears to have changed to use a copy tree, while the cleanup function `clean_outdir` was never updated to handle removing directories.

This bug is particularly egregious as packaging the project creates a `__pycache__` dir for every lambda, so rebuilding without running `cirrus clean` first is broken. That error looks something like this:

IsADirectoryError: [Errno 21] Is a directory: '***/.cirrus/lambdas/api/__pycache__'

tl;dr: `clean_outdir` needs to check for dirs and rmtree them. If encountering this issue run `cirrus clear` and try the build again.

jkeifer:
Closes #105.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+
+## [v0.6.0] - 2022-02-18
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


trevorskaggs (approved):
Pushed a minor linting change, otherwise :shipit:

PR #126, opened 2022-03-08
Additional slimPatterns for base config

sjwoodr:
During a recent deployment, it was found that the lambda deployment package uploaded to S3 had an unzipped size in excess of 95MB, which was rejected by AWS as too large.  The solution is to slim the numpy.libs and rasterio.libs directories and associated libraries.  The resulting zip file uploaded to the serverless deployment bucket was under 2MB (zipped) after this change.


jkeifer:
Seems like a reasonable change. I'll update the test fixtures when I have an opportunity, to get the tests passing, and merge this once that is done.

jkeifer:
So, it turns out we don't need these changes anymore as we have removed the offending lambdas from the default set deployed with cirrus. That said, we will need to keep these additional filter patterns in mind if we ever decide to distribute them in a cirrus plugin.

PR #127, opened 2022-03-11
workflow callbacks

jkeifer:
See cirrus-geo/cirrus-lib#46 for the lib side, which these changes depend on, as well as a lengthy write up regarding the workflow callbacks idea.


jkeifer:
I have decided to close this PR. The advent of the workflow notification topic has changed the prospective design of this feature, for the better. A new issue will be opened detailing the sketch of that idea, documenting the new vision. Eventually a new PR will be opened with an implementation of the idea to make callbacks a supported feature of cirrus.

PR #136, opened 2022-05-09
update-state support payload URLs

### LINKED ISSUE #135: `update-state` does not handle events with S3 URL payloads
In certain cases (typically failures), the `update-state` lambda can receive an event where the input and/or output payloads are a URL to an S3 object. The `ProcessPayload.from_event()` method will handle a payload URL, downloading and parsing the json before attempting initialization, but the `update-state` code is not using that constructor.

`update-state` should use the `from_event()` constructor in `parse_event()`. It also does not provide a traceback in the event of an exception; the logging call should be changed from `error()` to `exception()` to include that additional detail.

jkeifer:
Closes #135.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 
 - rerun feeder has required permissions ([#131])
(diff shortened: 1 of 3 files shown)


matthewhanson:
@jkeifer looks good, except needs a CHANGELOG addition

PR #141, opened 2022-05-31
fix docs build issue

jkeifer:
First problem: we were not installing cirrus-docs as a docs build dependency in the workflow. Second: cirrus-docs requires a specific version of cirrus, so we can't just put it in the requirements file and be done with it.

Lastly, the workflow wasn't triggered on PRs, which is why this wasn't caught before merge. I made the git commit/push conditional workflow steps, so the build will run on PRs and test that the docs are still good.


jkeifer:
Looks like we can't use the sphinx action because we don't currently support passing args to the executed sphinx command. I want to redo the way the command business works, such that the tool uses a build script and passes in env vars to that script, which could include an extra args string or something of that nature.

But for now, I'm just going to pull the sphinx action out and use a regular scripting action. I think the only difference is that will fail solely bases on sphinx return codes, rather than some newfangled output parsing, but I am not 100% sure the action doesn't do anything else desirable.

matthewhanson (approved):

PR #148, opened 2022-08-03
IAM permissions for "update-state" to read payloads from s3

ircwaves:
This is the IAM change that goes with the Python change in the `update-state` function.


ircwaves:
@jkeifer I think this should be good now.  I have tested in a deployment with events that are `{"url": "s3://{SNIPPED}/{UUID}.json"}` and all was well.

jkeifer (approved):
Looks good!

PR #156, opened 2022-09-29
support events from SNS

ircwaves:
PR as a fix or discussion starter:  

Not sure how often this comes up, but if a complicated workflow requires a large set of input Items, it can exceed the SNS message size limit (256kB as of this moment).  

Seems natural to use the handy uploading that `ProcessPayload(collection).get_payload` provides.  But that doesn't work with the current process lambda: 

https://github.com/cirrus-geo/cirrus-geo/blob/e07e899a82544cbad9711ec086db8ede45299ada/src/cirrus/builtins/functions/process/lambda_function.py#L59

which expects a ProcessPayload, and not an event.


ircwaves:
Given other development in process.  This is not needed.

Issue #120, opened 2022-02-21
`rerun` feeder should resolve full payloads, publish to process topic one-by-one


jkeifer:
Having the `process` lambda accept a list of payload IDs to resolve into full payloads has a few faults:

* If the payload ID is is not valid or some other issue happens when resolving the payload, it can affect the `process` operation for all other payload in the request and/or any other payloads batched up by SQS. Better error handling in `process` would help, but moving this payload resolution into the feeder is a better separation of concerns and simplifies the critical `process` logic (something also identified as a benefit in #103, which requires these `rerun` changes as a prerequisite).
* If the list of payload IDs is beyond a certain size, the expected runtime of the lambda cannot be known, which means we end up pushing the lambda timeout to unreasonable sizes (currently 900 seconds). With the current redrive policy of 5 attempts and our 1000 second visibility timeout to accommodate that lambda timeout, it takes 5000 seconds for unprocessable messages to fall into the DLQ, which is outside the bounds of SLAs for some use cases.

If we move the payload resolution into the `rerun` feeder itself we can solve both of these problems. `rerun` should have a way to handle payload resolution errors (or at least that they have happened in rerun means it is easier to know to look in its logs), and once it is resolving full payloads we should dial the timeouts back to a more reasonable setting (15s lambda timeout, 60s visibility timeout?).

jkeifer:
Closed by #157.

PR #157, opened 2022-10-11
`process` lambda cleanup and testing

### LINKED ISSUE #103: Cleanup/simplify process lambda
The process lambda supports a number of input payload types that complicate it, perhaps unnecessarily. We need to audit all the supported modes of operation and determine based on known use cases which we should preserve, then clean it up correspondingly.

(I believe, based on previous discussion, we only want to support a standard Cirrus ProcessPayload as input, either bare or from SQS, and support for any other input payload formats should be removed.)

jkeifer:
Closes #103.

Removes all the legacy undocumented code from process, and significantly reworks payload handling to allow partial success. This prevents errors with one payload tainting all others in a batch, such that all will eventually end up in the DLQ if only one is invalid.

The restructuring will also allow us to much more easily add metrics to the lambda, including exposing things like skipped and dropped payloads, which up to now have not been visible save for in the logs.

With the legacy features removed, we can also set the timeout to a reasonable value (18 sec, which still might be high), and adjust the SQS visibility timeout to match (20 sec).

To allow testing lambda handlers like this, I added a selection of utility functions which ultimately allow one to reference a lambda feeder/function/task by name and call the handler function as a regular python function. That allowed adding a full suite of tests against the `process` lambda. This is a _giant_ step forward for our test coverage, which up to now has simply covered the cli--actual core cirrus code has remained untested via automated means and has required deployment to AWS to be exercised manually. Now we just need to cover the rest of the builtins 😄.

Note that this PR breaks the rerun feeder, which will need to be fixed via #120.

Tests will be broken until we have a cirrus-lib release with the required changes and pin it on this branch. See cirrus-geo/cirrus-lib#56.


src/cirrus/builtins/functions/process/lambda_function.py
```diff
@@ -1,65 +1,88 @@
...
 
-# Default PROCESSES
-# TODO: put this configuration into the cirrus.yml
-with open(os.path.join(os.path.dirname(__file__), 'processes.json')) as f:
-    PROCESSES = json.loads(f.read())
+logger = get_task_logger('function.process', payload=tuple())
+
+
+def is_sqs_message(message):
+    return (
```
philvarner, line 13:
```suggestion
    return message.get('eventSource') == 'aws:sqs'
```

src/cirrus/builtins/functions/process/lambda_function.py
```diff
@@ -1,65 +1,88 @@
...
+                messages[payload['id']].append(message)
+            except KeyError:
+                messages[payload['id']] = [message]
 
+    processed_ids = set()
     if len(payloads) > 0:
-        _payloads = ProcessPayloads(payloads)
-        _payloads.process()
+        processed = ProcessPayloads(payloads).process()
+        logger.error(processed)
```
philvarner, line 57: is error the right level for these? I think it would also help here if there explanatory text in the log message for what was being logged, rather than just dumping the objects into the log

src/cirrus/builtins/functions/process/lambda_function.py
```diff
@@ -1,65 +1,88 @@
...
+
+    if failures:
+        # If we have partial failure, then we want to delete all
+        # successfully processed messages from the queue, so they
+        # won't be reprocessed again. We don't need to do this if
+        # we have no failures, as SQS will delete the messages for
+        # us if we exit successfully.
+        for message in successful_sqs_messages:
+            try:
+                utils.delete_from_queue(message)
```
philvarner, line 79: preferable to use [delete_message_batch()](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sqs.html#SQS.Client.delete_message_batch) here (in batches of 10) rather than one at a time, which can slow this down

src/cirrus/builtins/functions/process/lambda_function.py
jkeifer: Ha, that was terrible code. Thanks for the much better rewrite.

src/cirrus/builtins/functions/process/lambda_function.py
jkeifer: No, this was me being fast and sloppy with debug logging. Thanks for noticing.

src/cirrus/builtins/functions/process/lambda_function.py
jkeifer: I considered that initially, but it seemed like premature optimization. That is, chunking the successful messages into blocks of 10 adds complexity, and I don't think we'd often be failing here, so this won't run very often.

That said, you are probably right that we should do it anyway. As I reflect on it more, it seems like we either have only a handful of messages to process at once and they'll all fit into one batch delete call, or we'll have a bunch of messages and need to be using the batch call for performance reasons.

I will put it on the todo list to circle back around on this.

src/cirrus/builtins/cloudformation/resources.yml
```diff
@@ -19,7 +19,7 @@ Resources:
     Type: AWS::SQS::Queue
     Properties:
       QueueName: "#{AWS::StackName}-process"
-      VisibilityTimeout: 1000
+      VisibilityTimeout: 20
```
jkeifer, line 22: Found something interesting about this setting while looking for something else in the AWS docs: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#events-sqs-queueconfig.

Appears they recommend we set this to 6x the lambda timeout, which is 18s. So this should be 90s? I suppose that doesn't fail quite as fast as I hoped (for DLQ alerts), but it still isn't too bad.

jkeifer:
Note that the tests pass as we decided to move the relevant cirrus-lib files into this repo. We are moving towards deprecation of cirrus-lib, and want to move all relevant files into this project or replace them with the upcoming stac-task project.

As a result, this PR is no longer dependent on https://github.com/cirrus-geo/cirrus-lib/pull/56, which I have closed.

src/cirrus/lib2/process_payload.py
```diff
@@ -0,0 +1,384 @@
...
+        """Initialize a ProcessPayload, verify required fields, and assign an ID
+
+        Args:
+            state_item (Dict, optional): Dictionary of entry in StateDB. Defaults to None.
+        """
+        super().__init__(*args, **kwargs)
+
+        self.logger = get_task_logger(__name__, payload=self)
+
+        if self['type'] != 'FeatureCollection':
```
jkeifer, line 61: Don't need

src/cirrus/lib2/process_payload.py
```diff
@@ -0,0 +1,384 @@
...
+        self.tasks = self.process['tasks']
+
+        if 'workflow-' not in self['id']:
+            raise ValueError(f'Invalid payload id: {self["id"]}')
+
+        # TODO - validate with a JSON schema
+        #if schema:
+        #    pass
+        # For now, just make check that there is at least one item
+        if len(self['features']) < 1:
```
jkeifer, line 113: Remove this

src/cirrus/lib2/process_payload.py
```diff
@@ -0,0 +1,384 @@
...
+            raise ValueError(
+                'ProcessPayload.features is empty; must have at least one feature'
+            )
+
+        for item in self['features']:
```
jkeifer, line 118: Make all access of features use a get instead of keyword

src/cirrus/lib2/process_payload.py
```diff
@@ -0,0 +1,384 @@
...
+            if 'chain_filter' in process:
+                jsonfilter = jsonpath.parse(
+                    f'$.features[?({process["chain_filter"]})]',
+                )
+                new['features'] = [
+                    match.value for match in jsonfilter.find(new)
+                ]
+            yield new
+
+    def update(self):
```
jkeifer, line 171: This should be `create_id`? Should probably exit before doing all this stuff if `id` set.

This should raise an error if no `id` and no `features` or something else goes wrong making the `id`.

src/cirrus/lib2/utils.py
```diff
@@ -0,0 +1,187 @@
...
+        same = False
+
+    else:
+        # base case d1 == d2
+        pass
+
+    return same
+
+
+def extract_record(record):
```
jkeifer, line 109: Add test to for message written directly to SQS without SNS.

jkeifer:
I wrapped the rerun changes in here, minus tests. We'll want to get those added at some point, but they are not mission critical at the moment. If they don't get in before this is merged we should make a follow-up ticket for that (which we should do for all built-in lambdas anyway).

Thus, this closes #120 (though rerun just resolves payload URLs, not the full payload).

One blocker for merge is that cirrus build is not injecting cirrus.lib2 into these reworked lambdas. We'll need some way to change that from using cirrus.lib to cirrus.lib2 for those.

src/cirrus/builtins/functions/process/lambda_function.py
jkeifer: I have updated this with a batch method.

jkeifer:
Also needs updates to the CHANGELOG.

src/cirrus/builtins/cloudformation/resources.yml
jkeifer: I don't know where 90s came from. It should be 108s.

That makes the time from message receipt to a message falling into the DLQ roughly 9 minutes. Add additional time for alerting to fire, where applicable, and this is quite a long time.

I think we should also reduce the timeout on the `process` lambda itself to help get this time down. A value of 10s, for example make this visibility timeout 60s, providing a time-to-first-alert of 5-6 minutes, which I think is much more reasonable in cases with short SLAs. Perhaps I will go with these values for now but also make a follow up ticket to benchmark or analyze past `process` timing, to see if we should recommend reducing this value even further.

jkeifer: Created https://github.com/cirrus-geo/cirrus-geo/issues/167 as suggested.

jkeifer:
To get this moving forward and allow it to be used as a base for additional development, I am going to make a ticket to address the test coverage issue. I backed out the addition of the `__init__.py` files in the builtins tree.

We will have to make sure the relevant `CHANGELOG` entries are added before cutting the next release.

matthewhanson (approved):

PR #160, opened 2022-10-14
Handling large payloads

ircwaves:
1. Supports s3-url event from feeder.
2. Handle large payloads returned from `post-batch` and `publish` lambdas

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 __pycache__/
+.python-version
 .pytest_cache/
 node_modules/
 .idea/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Avoid `States.DataLimitExceeded` error when `publish` or `post-batch` return
+  large payloads by returning `payload.get_payload()`.
+
+### Added
+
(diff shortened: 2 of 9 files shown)


jkeifer:
I hate the name of `get_payload()`, it seems misleading and inaccurate. But whatever, let's ignore that for now.

So first thing, see #157. `process` has been completely refactored and supports payloads with URLs now. So I'd remove the changes to `process` here.

For the other tasks, we are likely moving to stac-task class-based tasks for these. The timeline on that change is uncertain, however, so I'm okay with these changes as an intermediate solution. Hence why I am ignore the name of `get_payload()`, because I am hoping its use will be short-lived. This said, I am not sure how we are solving this in stac-task, as this is something AWS-specific. @matthewhanson?

More generally, I'm still not sure what the best solution here is long-term. This whole message length limitation is a total pain and AWS should solve it for us, but obviously we can't rely on that and we need to do something...

I have some long-term ideas for cirrus that are coming out of a discussion with @matthewhanson about this new concept of a "stac-workflow" (which is made up of...stac-tasks, obviously). If that pans out, we'd have enough abstraction in the cirrus workflow (yes, I mean "the", not "a", as I envision a single dynamic workflow that can process anything) that I think switching to S3 for everything all the time might make the most sense. But the actual mechanics there are quite hand-wavy, so I don't want to commit to that trajectory.

tl;dr: I'm good with these changes if we pull out the `process` ones as an interim solution while we figure out what we really want to do.

jkeifer (approved):

Issue #158, opened 2022-10-12
use pre-commit to run black, flake8, etc


philvarner:
This is a good example of a pre-commit file:

https://github.com/stac-utils/stac-api-validator/blob/main/.pre-commit-config.yaml

and that project also has good config files for all of those tools, from the hypermodern python cookiecutter template.

jkeifer:
Closed by #171.

PR #171, opened 2022-10-24
linting and formatting fixes per pre-commit hooks

jkeifer:
Applies the pre-commit hooks to the project.

I have audited the changes as best I can and I don't see anything overly concerning (besides using `"` where `'` are so much better 😉). I think this is good to go, but I am hoping someone else can skim through and double check nothing stands out to them.

Re #158.


.isort.cfg
```diff
@@ -0,0 +1,2 @@
+[settings]
+profile=black
```
philvarner, line 2: sorry, i missed adding this file

philvarner (approved):

PR #175, opened 2022-10-25
add AWS Timestream event db for workflow state change events

philvarner:


requirements.txt
```diff
@@ -4,7 +4,8 @@ click-plugins~=1.1
 rich~=10.6
 cfn-flip~=1.2
 cirrus-lib~=0.8.0
-
 boto3-utils~=0.3.2
+boto3
```
philvarner, line 8: boto3 was already used, but not explicitly imported

requirements.txt
```diff
@@ -4,7 +4,8 @@ click-plugins~=1.1
...
 python-json-logger~=2.0
 jsonpath-ng>=1.5.3
+python-dateutil~=2.8.2
```
philvarner, line 11: this is a new dependency for it's ISO 8601 datetime parser.

src/cirrus/lib2/eventdb.py
```diff
@@ -0,0 +1,132 @@
...
+
+        event_time_dt = datetime.fromisoformat(event_time)
+        event_time_ms = str(int(event_time_dt.timestamp() * 1000))
+        duration_ms = str(
+            int((isoparse(last_update_ts_str) - event_time_dt).microseconds / 1000)
+        )
+
+        record = {
+            "Dimensions": [
+                {"Name": "collections_workflow", "Value": collections_workflow},
```
jkeifer, line 59: Could we split this to be `workflow` as the first dimension and `collections` as the second?

philvarner: yes

src/cirrus/lib2/statedb.py
```diff
@@ -135,7 +143,7 @@ def get_items_page(
         limit=100,
         nextkey: str = None,
         **kwargs,
-    ) -> List[Dict]:
```
philvarner, line 138: this type annotation was wrong

src/cirrus/lib2/eventdb.py
```diff
@@ -0,0 +1,144 @@
...
+    ABORTED = "ABORTED"
+
+
+class EventDB:
+    def __init__(self, event_db_and_table_names: Optional[str] = None):
+        self.tsw_client = boto3.client("timestream-write")
+        self.tsq_client = boto3.client("timestream-query")
+
+        if event_db_and_table_names is None:
+            event_db_and_table_names = os.getenv("CIRRUS_EVENT_DB_AND_TABLE")
```
jkeifer, line 27: It seems like this needs to be added to the default `cirrus.yml` file. Look at `src/cirrus/core/config/default.yml`.

src/cirrus/lib2/eventdb.py
```diff
@@ -0,0 +1,144 @@
...
+    def write_timeseries_record(
+        self,
+        key: Dict[str, str],
+        state: StateEnum,
+        event_time: str,
+        execution_arn: Optional[str] = None,
+    ) -> Any:
+        parts = key.get("collections_workflow", "").rsplit("_", 1)
+        if not len(parts) == 2:
+            logger.error(
```
jkeifer, line 43: Is there a reason to simply log an error vs raising an exception? It seems like it should be up to the caller to determine how to handle a failure case.

Same goes for other places with similar log and return behavior.

src/cirrus/lib2/statedb.py
```diff
@@ -343,7 +350,10 @@ def set_completed(
             UpdateExpression=expr,
             ExpressionAttributeValues=expr_attrs,
         )
-        logger.debug("set outputs", extra=key.update({"outputs": outputs}))
+        logger.debug("set completed", extra=key.update({"outputs": outputs}))
+
+        self.write_timeseries_record(key, StateEnum.COMPLETED, now)
```
jkeifer, line 355: I feel rather strongly we should make writing the timeseries record conditional on having the execution ARN. If the barrier here is simply passing in that in from `update-state`, I am happy to throw a commit on here to plumb the ARN though.

To recap my concern, we will not be able to group the terminal events for an execution with its start, so we will not be able to calculate things like duration after the fact with certainty that we've related the correct events together.

Another major reason that has solidified my opinion here is using these methods in some ad hoc management scripts I threw together to manage some issues with missed state updates on a recent large ingest. These sorts of things can happen for several reasons when processing millions of items, so being able to mitigate the problems with custom scripts that reuse these methods is super helpful. But in such cases I probably don't want events.

tests/builtins/functions/test_process.py
```diff
@@ -29,6 +29,7 @@ def add_event_info(_dict):
 @pytest.fixture(autouse=True)
 def process_env(workflow):
     workflow_prefix = workflow["stateMachineArn"].rsplit(":", 1)[0] + ":"
+    os.environ["CIRRUS_EVENT_DB_AND_TABLE"] = "event-db-1|event-table-1"
```
jkeifer, line 32: I don't think you need this here if it is in the builtins `env` fixture.

src/cirrus/lib2/eventdb.py
philvarner: I tend to prefer a style where Exceptions aren't thrown. In this case, if something is wrong with the parameters provided to the eventdb, I don't want it to possibly propagate an error out and cause the workflow to fail -- I'd rather miss the event and let the workflow succeed.

tests/builtins/functions/test_process.py
philvarner: yeah, this was leftover from when I was trying to figure out why the env var wasn't being propagated correctly.  removed.

src/cirrus/lib2/eventdb.py
philvarner: Well, now instead of just logging the failures in the the statedb tests, they fail, so I'm going to have to figure that out out lol. (I probably just need to mock in the service with moto)

philvarner: ValueErrors added, tests fixed

src/cirrus/builtins/cloudformation/resources.yml
```diff
@@ -61,6 +61,20 @@ Resources:
...
+      # KmsKeyId: String -- use default one
+  StateEventTimestreamTable:
+    Type: AWS::Timestream::Table
+    DependsOn: StateEventTimestreamDatabase
+    Properties:
+      TableName: "#{AWS::StackName}-state-events-table"
+      DatabaseName: "#{AWS::StackName}-state-events"
+      RetentionProperties:
+        MemoryStoreRetentionPeriodInHours: "24"
+        MagneticStoreRetentionPeriodInDays: "93"
```
philvarner, line 77: @jkeifer changed this to 93 from 3650 (10 years), for at least 2 full months of data. Keeping this around forever is probably unnecessary, and users can change it if they want.

src/cirrus/builtins/functions/update-state/lambda_function.py
```diff
@@ -40,67 +42,130 @@
...
-    statedb.set_aborted(input_payload["id"])
+def workflow_aborted(execution: Execution) -> None:
+    statedb.set_aborted(execution.input["id"], execution_arn=execution.arn)
 
 
-def workflow_failed(input_payload, output_payload, error):
-    # error type
-    error_type = error.get("Error", "unknown")
+def workflow_failed(execution: Execution) -> None:
+    error_type = "unknown"
```
philvarner, line 135: @jkeifer reworked this a bit since execution.error may be None

tests/builtins/conftest.py
```diff
@@ -71,12 +65,6 @@ def queue(sqs):
     return q
 
 
-@pytest.fixture
-def statedb(dynamo, statedb_schema):
-    dynamo.create_table(**statedb_schema)
-    return statedb_schema["TableName"]
-
-
```
jkeifer, line 79: This doesn't seem right...I replaced the statedb fixture, but it should still be required here. See https://github.com/cirrus-geo/cirrus-geo/blob/a9e25850c29c2647759ac8ca922083f3c32fc6f3/tests/builtins/conftest.py#L70.

Maybe a merge conflict?

This also means the dynamo mock above shouldn't be removed.

jkeifer: Never mind, the file sorting confused me when I was looking for the top-level conftest.py and I didn't see it. Ignore. :)

jkeifer (approved):

Issue #173, opened 2022-10-25
API/rerun fail with dynamo client error


jkeifer:
We get an error like `ClientError: An error occurred (ValidationException) when calling the Query operation: The provided starting key is invalid`. Looks like something related to the exclusive start key.

PR #176, opened 2022-10-25
`api` and `rerun` fixes

jkeifer:
Relates to #173 and #120, mainly bugfixes but also some cleanup around logging.

### DIFF
--- a/src/cirrus/builtins/feeders/feed-rerun/lambda_function.py
+++ b/src/cirrus/builtins/feeders/feed-rerun/lambda_function.py
@@ -6,8 +6,11 @@ from os import getenv
 
 import boto3
 
+from cirrus.lib2.logging import defer, get_task_logger
 from cirrus.lib2.statedb import StateDB
 from cirrus.lib2.utils import batch_handler, submit_batch_job
+
+logger = get_task_logger("feeder.rerun", payload=tuple())
 
 # envvars
 SNS_TOPIC = getenv("CIRRUS_PROCESS_TOPIC_ARN")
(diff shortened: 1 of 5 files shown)


src/cirrus/builtins/feeders/feed-rerun/lambda_function.py
```diff
@@ -18,22 +21,25 @@
...
 def submit(payload_ids):
-    with batch_handler(
-        SNS_CLIENT.publish_batch,
-        {"TopicArn": SNS_TOPIC},
-        "PublishBatchRequestEntries",
-    ) as handler:
+    with batch_handler(publish_batch, {}, "messages", batch_size=10) as handler:
         for payload_id in payload_ids:
-            handler.add({"url": StateDB.get_payload_url(payload_id)})
+            handler.add(json.dumps({"url": StateDB.get_payload_url(payload_id)}))
```
jkeifer, line 38: In running this I had an error because I wasn't constructing the entries properly (they needed to be dicts with `Id` and `Message` keys. I created the `publish_batch` function to handle building those entries.

For background, the `batch_handler` is a context manager factory function that returns a `BatchHandler` class (https://github.com/cirrus-geo/cirrus-geo/blob/main/src/cirrus/lib2/utils.py#L252), which runs the provided function for every `batch_size` items added to the handler. In the case that the batch size doesn't divide evenly into the number of items, the context manager will ensure any remaining items added but not processed will be flushed to the specified function at context exit.

I'm also using this in the `process` lambda to handle the batching of SQS message deletes in the case of partial success.

src/cirrus/builtins/functions/api/lambda_function.py
```diff
@@ -1,14 +1,14 @@
...
-import logging
 import os
 from copy import deepcopy
 from urllib.parse import urljoin
 
 from boto3utils import s3
 
-from cirrus.lib.statedb import STATES, StateDB
+from cirrus.lib2.logging import get_task_logger
+from cirrus.lib2.statedb import STATES, StateDB
```
jkeifer, line 9: `lib2.statedb` has the bugfix.

src/cirrus/builtins/feeders/feed-rerun/lambda_function.py
```diff
@@ -6,9 +6,12 @@
 
 import boto3
 
+from cirrus.lib2.logging import defer, get_task_logger
 from cirrus.lib2.statedb import StateDB
```
jkeifer, line 10: `lib2.statedb` has the bugfix.

src/cirrus/lib2/logging.py
```diff
@@ -30,16 +30,15 @@
...
             "handlers": ["standard"],
             "level": getenv("CIRRUS_LOG_LEVEL", "DEBUG"),
         },
     },
 }
 
 
-def configure_logging():
-    logging.config.dictConfig(config)
+logging.config.dictConfig(config)
```
jkeifer, line 41: Moving this into the function so it didn't run at module import was done while testing an issue that popped up when this code was migrated from cirrus-lib. I ended up solving the problem a different way, but this change inadvertently remained. Moving the logging config to the module scope ensures logging is configured correctly by default.

src/cirrus/lib2/statedb.py
```diff
@@ -490,14 +490,15 @@ def query(
             index = "state_updated"
             expr = expr & Key(index).begins_with(state)
 
-        keys = ["collections_workflow", "itemids"]
+        exclusive_start_key_filters = ["collections_workflow", "itemids"]
 
         if index:
             kwargs["IndexName"] = index
+            exclusive_start_key_filters.append(index)
```
jkeifer, line 497: This was the underlying issue affecting `api` and `rerun`. In short, an older `moto` version broke the dynamo tests when `index` was included in the `ExclusiveStartKey`, so I ended up removing this line to get those tests to work because I thought this was a bug in the code here. Turns out it was a bug in `moto`; a newer moto version fixes this (though still has different behavior than dynamo, in that omitting `index` still works with `moto` but not with dynamo).

Adding this line back in resolves the problem. All other changes in this file are just for clarity/logging.

philvarner (approved):

Issue #177, opened 2022-10-27
All builtins need to be updated to use lib2 and/or stac-task


jkeifer:
We are moving away from the cirrus-lib dependency and instead packaging core functionality into cirrus-geo under `cirrus.lib2` (the name is perhaps temporary; once we can be sure everything is off cirrus-lib and we are no longer packaging it into lambda tasks we can rename to `cirrus.lib` if desired). Some of the critical builtins have been updated to use lib2 as necessary, but others are still using cirrus-lib's `cirrus.lib`.

Where possible, we should update tasks to depend on stac-task instead of `cirrus.lib` or `cirrus.lib2`. For cases where `cirrus.lib` is still required we need to update them to use `cirrus.lib2` or otherwise extract the relevant lib code into the task (if it is not shared, for example `publish`).

Once this is done we can stop packaging lib/lib2 by default and instead pull lib2 in only where necessary.

PR #179, opened 2022-10-27
Update remaining builtins to use `cirrus.lib2`

### LINKED ISSUE #170: Update `update-state` validation to match `process`
Recent refactor of the `process` lambda removed several restrictions around the payload structure, requiring certain fields (e.g., FeatureCollection type, features array) even though the payload isn't necessarily valid GeoJSON or STAC.

However, the `update-state` lambda still uses the old validation scheme in cirrus-lib ProcessPayload, so a new payload taking advantage of the relaxed restrictions that fails, can cause update-state to fail parsing the payload and the state gets stuck in PROCESSING

jkeifer:
Closes #170 and #177.

Note that I did not use stac-task, as some of the builtins work a bit lower-level, and that wasn't going to provide the right level of abstraction. Sticking with lib2 seemed most appropriate/easiest for the time being.

I also have not tested these changes. They are minimal and they should work, but we'll want to deploy to a real environment and see that they all work before cutting a release (and someday we'll add automated testing for all of them so we have confidence in such changes).

### DIFF
--- a/src/cirrus/builtins/tasks/publish/lambda_function.py
+++ b/src/cirrus/builtins/tasks/publish/lambda_function.py
@@ -1,18 +1,103 @@
 import json
 from os import getenv
 
-from cirrus.lib.logging import get_task_logger
-from cirrus.lib.process_payload import ProcessPayload
-from cirrus.lib.statedb import StateDB
+import boto3
+
+from cirrus.lib2.logging import get_task_logger
+from cirrus.lib2.process_payload import ProcessPayload
+from cirrus.lib2.statedb import StateDB
 
 # envvars
 DATA_BUCKET = getenv("CIRRUS_DATA_BUCKET")
 API_URL = getenv("CIRRUS_API_URL", None)
+PUBLISH_TOPIC_ARN = getenv("CIRRUS_PUBLISH_TOPIC_ARN", None)
 # DEPRECATED - additional topics
 PUBLISH_TOPICS = getenv("CIRRUS_PUBLISH_SNS", None)
 
 # Cirrus state db
 statedb = StateDB()
+snsclient = boto3.client("sns")
+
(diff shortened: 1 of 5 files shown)


src/cirrus/builtins/tasks/publish/lambda_function.py
```diff
@@ -1,18 +1,103 @@
...
+        }
+
+    if "datetime" in item["properties"]:
+        attr["datetime"] = {
+            "DataType": "String",
+            "StringValue": item["properties"]["datetime"],
+        }
+
+    if "eo:cloud_cover" in item["properties"]:
+        attr["cloud_cover"] = {
```
philvarner, line 71: I know this was copied from somewhere else, but I was wondering if it was intentional that the eo:cloud_cover field name is changed to cloud_cover in the event?

jkeifer: The name change is because SNS does not support : in attribute names. I’m not sure if dropping the eo is almost appropriate here, but I’d rather not make this a breaking change unless we could preserve the entire attribute name.

src/cirrus/builtins/tasks/publish/lambda_function.py
```diff
@@ -1,18 +1,103 @@
...
 # DEPRECATED - additional topics
 PUBLISH_TOPICS = getenv("CIRRUS_PUBLISH_SNS", None)
 
 # Cirrus state db
 statedb = StateDB()
+snsclient = boto3.client("sns")
+
+
+def sns_attributes(item) -> dict:
+    """Create attributes from Item for publishing to SNS
```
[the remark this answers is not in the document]
jkeifer: Good point. I will update with some additional conditionals.

PR #198, opened 2022-11-05
drop support for python 3.7

philvarner:
Drops support for Python 3.7.

- Add a python version restriction in the setup.py
- Removes 3.7 from the GH testing matrix, and adds 3.11 (even though this version is not supported in AWS Lambda yet, as 3.10 also is not, but it allows us to catch issues with newer versions earlier)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -66,6 +66,7 @@ dependency_links = [x.strip().replace("git+", "") for x in reqs if "git+" not in
 
 setup(
     name="cirrus-geo",
+    python_requires=">=3.8",
     packages=find_namespace_packages("src"),
     package_dir={"": "src"},
     version=VERSION,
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -66,6 +66,7 @@ def git_version(gitdir, default="0.0.0"):
 
 setup(
     name="cirrus-geo",
+    python_requires=">=3.8,<=3.9",
```
jkeifer, line 69: Is there a good reason to set an upper limit here?

For one, once 3.10 is available as a runtime, it seems like also forcing a cirrus-geo upgrade to make use of it despite the code being compatible is not great. For another, you can technically run the cirrus cli using a more recent python version; serverless packaging the lambdas will use whatever version of python is specified by the runtime parameter in the cirrus.yml, which can be a different version (though that is not recommended).

Because of those points I'd suggest not placing an upper bound, but I am open to opinions.

jkeifer (approved):

setup.py
philvarner: I had originially put it at <=3.9 b/c we haven't system tested it with 3.10, but then I realized that doesn't work for restricting it to any 3.9.x version, and it should be `<3.10` instead -- but I also don't thing the restriction matters that much in practice, so I removed it.

PR #201, opened 2022-11-06
Changes to support new management cli plugin

### LINKED ISSUE #121: Commands to manage cirrus pipeline
A cirrus project should have sufficient information available to be able to glean all the configured resources in AWS (perhaps using cloudformation describe-stacks to get the stack outputs, or getting the process lambda configuration ENV var definitions) such that it could provide cli commands to interact with and manage that pipeline deployment. An MVP for this command set should include:

* querying recent executions a la the API with filtering for certain states, etc.
* a way to pull a process payload by payload ID
* a way to enqueue payloads for processing

The discovered config should be cached locally per stage, and we should have some commands to administer that cached configuration state, such as dumping out, refreshing, or clearing those state caches.

jkeifer:
I have started a new management plugin: cirrus-geo/cirrus-mgmt. The plugin requires some supporting changes in this repo, which is what this PR is for. Between a release of that cli and merging this PR we can close #121.

### DIFF
--- a/src/cirrus/lib2/statedb.py
+++ b/src/cirrus/lib2/statedb.py
@@ -8,40 +8,51 @@ from boto3.dynamodb.conditions import Attr, Key
 
 from .eventdb import EventDB, StateEnum
 
-# envvars
-PAYLOAD_BUCKET = os.getenv("CIRRUS_PAYLOAD_BUCKET")
-
 STATES = ["PROCESSING", "COMPLETED", "FAILED", "INVALID", "ABORTED"]
 
 # logging
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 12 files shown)


src/cirrus/lib2/statedb.py
```diff
@@ -512,13 +518,38 @@ def query(
         if self.limit and ("Limit" not in kwargs or self.limit < kwargs["Limit"]):
             kwargs["Limit"] = self.limit
 
-        logger.debug(kwargs)
```
jkeifer, line 515: fix this

ircwaves (approved):
Bueno!

PR #210, opened 2023-01-18
update executions URL per AWS deprecation of old URL

ircwaves:
AWS announced deprecation of the current URL format cirrus uses, this switches over to the new scheme.


tests/cli/test_cli.py
```diff
@@ -32,6 +32,8 @@ def shasum(path: Path):
 def hash_tree(src_dir):
     hashed = {}
     for root, dirs, files in os.walk(src_dir):
+        # os.walk needs _in place_ filter of dot-dirs
+        dirs[:] = [d for d in dirs if not d.startswith(".")]
```
jkeifer, line 36: Ooh, that's tricky. Thanks for that syntactical tip!

jkeifer (approved):

tests/cli/test_cli.py
ircwaves: syntactic sugar is the sweetest.

PR #215, opened 2023-03-09
Fix doc builder and fix links on readme links

### LINKED ISSUE #212: docs build is broken
```
python3 update-versions.py gh-pages
Traceback (most recent call last):
  File "/home/runner/work/cirrus-geo/cirrus-geo/docs/update-versions.py", line 61, in <module>
    main(sys.argv[1])
  File "/home/runner/work/cirrus-geo/cirrus-geo/docs/update-versions.py", line 26, in main
    stable_version = parse("")
  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/packaging/version.py", line 52, in parse
    return Version(version)
  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/packaging/version.py", line 197, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: ''
make: *** [Makefile:[27](https://github.com/cirrus-geo/cirrus-geo/actions/runs/4017777368/jobs/6902565047#step:6:28): gh-pages-versions-update] Error 1
Error: Process completed with exit code 2.
```

tackaberry:
Fixes #212 

I was experiencing the same broken action on my fork. 
image

The changes fix the errors
image

Im not sure why this worked before but `parse()` was failing on strings like `""` and `main`. 

I also fixed some links on the readme - and adjust the text to match the documentation. 
image

### DIFF
--- a/docs/update-versions.py
+++ b/docs/update-versions.py
@@ -23,11 +23,12 @@ def main(gh_pages_dir):
     gh_pages_dir = Path(gh_pages_dir)
 
     all_versions = set()
-    stable_version = parse("")
+    stable_version = parse("0.0")
     for _file in gh_pages_dir.iterdir():
         if _file.is_dir():
-            version = parse(_file.name)
-            if not isinstance(version, Version):
+            try:
+                version = parse(_file.name)
+            except:
                 continue
             version.name = _file.name
             all_versions.add(version.name)
--- a/docs/update-versions.py
+++ b/docs/update-versions.py
@@ -23,7 +23,8 @@ def main(gh_pages_dir):
     gh_pages_dir = Path(gh_pages_dir)
 
     all_versions = set()
-    stable_version = parse("0.0")
+    default_version = parse("0.0")
+    stable_version = default_version
     for _file in gh_pages_dir.iterdir():
         if _file.is_dir():
             try:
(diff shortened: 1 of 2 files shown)


docs/update-versions.py
```diff
@@ -23,11 +23,12 @@ def main(gh_pages_dir):
     gh_pages_dir = Path(gh_pages_dir)
 
     all_versions = set()
-    stable_version = parse("")
+    stable_version = parse("0.0")
```
jkeifer, line 26: The reason this was parsing an empty string was to get a version instance with no value, such that it could be used in comparisons with other versions, but would also be falsy. Setting this to the parsed value of `"0.0"` will result in a version instance that would be truthy.

I think it would be best to do something like this so we can retain the default for comparison below:
```suggestion
    default_version = parse("0.0")
    stable_version = stable_version
```

jkeifer:
The reason that this broke is because the `packaging` module (https://pypi.org/project/packaging/) is pulled in by the `cirrus-docs` plugin, but not pinned in this project. A newer major version of that package broke compatibility with the version we were using, thus the docs build broke.

It might be worth strictly pinning that and any other dependencies required for the docs build, but I can also see value in the argument that keeping them floating will ensure we are getting the latest versions without having to manually update them. The downside to the latter approach is obviously this situation where things break and stay broken for some time.

docs/update-versions.py
tackaberry: Good poijnt @jkeifer, I like the suggestion. Fixed in new commit.

jkeifer (approved):

PR #219, opened 2023-07-14
fixes batch tasks using stac-task

matthewhanson:
This fixes batch tasks using the stac-task library.

The post-batch Lambda expects an output payload with the same URL as the input payload except ending in `_out.json` rather than `.json`.  However, stac-task requires specifying the output payload URL explicitly which was not possible in the JobDefinition of the CloudFormation files.

This PR has the pre-batch Lambda passing both a `url` and `url_out` parameter to the Batch task so that `url_out` can be passed to the stac-task CLI.  The post batch Lambda then looks at the parameters to get the `url_out` to read the payload from.

JobDefinition's using stac-task should be structured like:

```
    Parameters:
      url: ""
      url_out: ""
    ContainerProperties:
      Command:
        - task
        - run
        - Ref::url
        - --output
        - Ref::url_out
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,31 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [v0.9.1] - 2023-07-14
+
 ### Fixed
 
 - Include `requirements.txt` for install from source distribution. This was
   missing and prevented install from pypi source.
+- For Batch tasks in workflows, the output payload URL is now explicitly set in the pre-batch lambda
+so that the URL is in the Parameters list of the output, rather than the post-batch function
+having to infer the output payload URL.  This fixes Batch tasks when using stac-task.
+Any JobDefinition using stac-task should specify `url` and `url_out` as Parameters and specify --output
+in the Command, e.g.:
+
+```
+    Parameters:
+      url: ""
+      url_out: ""
+    ContainerProperties:
+      Command:
+        - task
+        - run
+        - Ref::url
+        - --output
+        - Ref::url_out
+```
+
 
 ## [v0.9.0] - 2023-01-26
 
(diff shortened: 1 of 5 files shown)


matthewhanson:
Looks like I got me some tests to update

src/cirrus/lib2/utils.py
```diff
@@ -168,9 +168,10 @@ def extract_record(record):
     if "Message" in record:
         record = json.loads(record["Message"])
 
-    if "url" not in record and "Parameters" in record and "url" in record["Parameters"]:
+    if "url" not in record and "Parameters" in record:
         # this is Batch, get the output payload
-        record = {"url": record["Parameters"]["url"].replace(".json", "_out.json")}
+        if "url_out" in record["Parameters"]:
+            record = {"url": record["Parameters"]["url_out"]}
```
ircwaves, line 174: So, I could be wrong here, but for the avoidance of a breaking change, I'd recommend issuing a deprecation warning, and supporting the legacy behavior
```suggestion
            record = {"url": record["Parameters"]["url_out"]}
        else:
            record = {"url": record["Parameters"]["url"].replace(".json", "_out.json")}
            import warnings
            warnings.warn("Using the deprecated implicit '_out.json' url-hack",
                          category=DeprecationWarning, stacklevel=2)
```

ircwaves (changes requested):

src/cirrus/lib2/utils.py
matthewhanson: I was wondering if I should have this be backwards compatible, but then thought that if you upgrade to this version you are updating both the pre and post batch lambdas...so is there a case where `url_out` would not be present?

ircwaves: Oh, I see. pre-batch uses the same WKloc, so it does work w/o changing the batch task, but the task should still be changed.

matthewhanson:
@ircwaves I think @philvarner is looking to release a 0.10.0 rather than a 0.9.1, at any rate that should probably be another PR (for bumping version)

ircwaves (approved):

CHANGELOG.md
```diff
@@ -617,7 +638,8 @@ cleanup steps.
 
 Initial release
 
-[unreleased]: https://github.com/cirrus-geo/cirrus-geo/compare/v0.9.0...main
+[unreleased]: https://github.com/cirrus-geo/cirrus-geo/compare/v0.9.1...main
+[v0.9.0]: https://github.com/cirrus-geo/cirrus-geo/compare/v0.9.0...v0.9.1
```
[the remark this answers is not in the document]
ircwaves: Per offline discussion, this is good to go.

PR #221, opened 2023-07-19
drop support for python 3.8, update boto3-utils to 0.4.1

philvarner:


matthewhanson (approved):

philvarner:
@matthewhanson boto3-utils 0.4.1 doesn't support 3.8 because it uses methods only available in 3.9, so are you okay with the change here to drop support for python 3.8 in cirrus-geo?

ircwaves (commented):
Excellent!

PR #230, opened 2023-12-05
purge old external cirrus.lib package

ircwaves:
With the archival of `cirrus-lib` it seems reasonable time for the external cirrus-lib package to go now, agreed?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Removed
+
+- Cleaned `cirrus.lib` (old separate package) from lambda templates and build
+  process. ([#230])
+
 ### Fixed
 
 - Ensure correct count returned from `process` lambda and resolve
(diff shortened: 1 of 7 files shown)


ircwaves:
@jkeifer -- I don't know if this change is too soon or not, but after bonking the "re-run jobs" button twice (only one job seems to be able to upload coverage at a time), the tests passed w/ the change in coverage likely due to winning at code-golf.

philvarner:
@jkeifer thoughts about merging this?

jkeifer:
I'm good with this, the only thing I'd call out is adding a clear "Breaking Changes" section to the `CHANGELOG.md` (as has been done in the past for breaking changes) with the instruction that any users relying on the automatic packaging of `cirrus.lib` into lambdas will need to explicitly add `cirrus.lib` to those function requirements (and that `cirrus.lib` is no longer maintained and that we recommend migrating to `stac-task`).

philvarner:
> I'm good with this, the only thing I'd call out is adding a clear "Breaking Changes" section to the `CHANGELOG.md` (as has been done in the past for breaking changes) with the instruction that any users relying on the automatic packaging of `cirrus.lib` into lambdas will need to explicitly add `cirrus.lib` to those function requirements (and that `cirrus.lib` is no longer maintained and that we recommend migrating to `stac-task`).

I'll make that change

philvarner (approved):

PR #231, opened 2024-01-05
improve post-batch error handling

philvarner:
Resolves a few issues:

1. If the job was never tried, there's no `Attempts` array, which throws:
```
"Cause": "{\"errorMessage\": \"list index out of range\", \"errorType\": \"IndexError\", \"requestId\": \"e0980b7b-27f9-4501-a7f0-c99e76fd81c3\", \"stackTrace\": [\"  File \\\"/var/task/lambda_function.py\\\", line 23, in lambda_handler\\n    logname = cause[\\\"Attempts\\\"][-1][\\\"Container\\\"][\\\"LogStreamName\\\"]\\n\"]}"
```

2. Failure to retrieve the cloudwatch logstream (which will happen if the stream was named but never logged to, so it doesn't exist) used `logger.exception("Original exception:")` missing the actual original *error* message, and the dynamodb failure message is just "Exception: Unable to get error log.". This now logs the Reason instead, which is typically due to a failure to run the Batch task at all.

3. Better handling for other cases where the message is incomplete.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Fixed
 
+- In post-batch, better handle errors with the task failing to run and/or the CloudWatch log
+  not existing.
 - Ensure correct count returned from `process` lambda and resolve
   `UnboundLocalError` encountered on certain workflow failures. ([#224])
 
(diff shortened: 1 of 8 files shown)


jkeifer (approved):
I think this is great! I'd love to see us translate the non-batch-job-emitted errors to unique exception classes that could be matched in step function catch blocks, but that's probably a lot more work and this work looks to be a good step to enabling that in the future.

PR #232, opened 2024-01-10
release prep 0.10.1

philvarner:


CHANGELOG.md
```diff
@@ -5,12 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [v0.10.1] - 2024-01-10
```
jkeifer, line 8: Looks like you are missing the link def for `[v0.10.1]`? Also, I tend to leave `## [Unreleased]` in even if it is empty so it doesn't have to get added back in later and you don't have to comment the link (just update it for the new base tag), but 🤷 #youdoyou 😄.

jkeifer (approved):

PR #240, opened 2024-02-02
Documentation clarification and more info on using Batch

philvarner:
Several additions to the docs, integrating what I learned from the Earth Search Sentinel-2 ingest work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,25 +4,28 @@
 
 # Cirrus
 
-Cirrus is a [STAC](https://stacspec.org/)-based processing pipeline.
-As input, Cirrus takes a GeoJSON FeatureCollection with 1 or more STAC Items
-(an `ItemCollection` in pystack nomeclature), with a process definition block.
-That input is called a cirrus `ProcessPayload` (CPP).
+Cirrus is a [STAC](https://stacspec.org/)-based geospatial processing pipeline platform,
+implemented using a scalable architecture deployed on AWS. Cirrus provides the generic
+infrastructure for processing, allowing a user to focus on implementing the specific
+processing logic for their data.
+
+![](src/cirrus/docs/images/arch-overview.png)
+
+As input, Cirrus takes a STAC ItemCollection, with a process definition block.
+That input is called a Cirrus `ProcessPayload` (CPP).
 
 An input is run through a workflow that generates one or more output STAC Items.
 These output Items are added to the Cirrus static STAC catalog in S3,
 and are also broadcast via an SNS topic. Subscrptions to that topic can
-triggering additional workflows or external processes, such as for keeping
-a dynamic STAC catalog up to date (for example,
-[STAC-server](https://github.com/stac-utils/stac-server)).
+triggering additional workflows or external processes, such as indexing into a
+STAC API catalog (e.g.,
+[stac-server](https://github.com/stac-utils/stac-server)).
 
 Cirrus workflows range from the simple publishing of unmodified input items to
 the complex transformation of input Items and generation of wholly-new output
 Items. The current state of CPP processing is tracked in a state database
 to prevent duplicate processing and allow for a user to follow the state of any
 input through the pipeline.
-
-![](docs/images/highlevel.png)
 
 As shown in this high-level overview of Cirrus, users input data to Cirrus
 through the use of _feeders_. Feeders are simply programs that get/generate
--- a/src/cirrus/docs/10_intro.rst
+++ b/src/cirrus/docs/10_intro.rst
@@ -93,7 +93,7 @@ accessible to end-users through the robust ecosystem of STAC tooling.
 
 
 Example Earth Search use case
-----------------
+-----------------------------
 
 One prominent use case of Cirrus is as the processing pipeline for `Earth Search`_.
 
@@ -115,7 +115,7 @@ API and ingesting any new scenes. Similarly, the older Sentinel-2 Level-2A colle
 is populated by receiving SNS messages from an older Cirrus pipeline that supported the
 now-deprecated Earth Search v0 deployment.
 
-.. _earth-search: https://www.element84.com/earth-search/
+.. _earth search: https://www.element84.com/earth-search/
 
 AWS services used
 -----------------
--- a/src/cirrus/docs/10_intro.rst
+++ b/src/cirrus/docs/10_intro.rst
@@ -9,7 +9,7 @@ to massive workloads of millions of items in both a cost-efficient and
 performance-efficient manner, regardless if processing takes
 seconds, hours, or longer.
 
-The core of Cirrus consists of two primary pieces:
+Cirrus consists of two primary pieces:
 - `cirrus`, a CLI-based project management and deploy tool, and
 - `cirrus-lib`, a Python library providing a number of useful
 abstractions solving common needs for users writing their own Cirrus components.
@@ -30,7 +30,7 @@ Concepts
 STAC-based workflows
 ^^^^^^^^^^^^^^^^^^^^
 
-A key design aspect of Cirrus is the use of the `STAC`_ metadata specification as a
+A core design aspect of Cirrus is the use of the `STAC`_ metadata specification as a
 basis for the :doc:`Cirrus Process Payload <30_payload>` format. In this
 way Cirrus, encourages a highly-interoperable, metadata-first approach for both
 pipeline operators and end users alike.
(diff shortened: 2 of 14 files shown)


README.md
```diff
@@ -4,26 +4,29 @@
...
+STAC API catalog (e.g.,
+[stac-server](https://github.com/stac-utils/stac-server)).
 
 Cirrus workflows range from the simple publishing of unmodified input items to
 the complex transformation of input Items and generation of wholly-new output
 Items. The current state of CPP processing is tracked in a state database
 to prevent duplicate processing and allow for a user to follow the state of any
 input through the pipeline.
 
-![](docs/images/highlevel.png)
```
philvarner, line 25: this image link was broken, so I replaced it with the arch overview and put it near the top.

src/cirrus/docs/10_intro.rst
```diff
@@ -29,35 +30,37 @@ Concepts
...
-  processing pipeline to transform a given input into one or more output STAC
-  items
-
-An additional component type is that of a :doc:`Function
-<components/functions>`, though they are less commonly extended by end users.
-
+<60_components>`. There are four types of Components:
+
+* :doc:`Feeders <components/feeders>` accept arbitrary input, and output a
+  Cirrus Process Payload, which is enqueue for processing. These can be used to initiate
```
stark525, line 52: enqueue -> enqueued?

src/cirrus/docs/10_intro.rst
```diff
@@ -83,34 +86,56 @@ Relationship with stac-server
...
 
 .. _stac-server: https://github.com/stac-utils/stac-server
 
 
-Example use cases
------------------
+Example Earth Search use case
+----------------
+
+One prominent use case of Cirrus is as the processing pipeline for `Earth Search`_.
```
stark525, line 98: Is it worth digging slightly deeper into how these map to cirrus concepts? maybe even a diagram of the specific Landsat workflow with USGS's SNS topic? (this can be a separate issue / improvement later)

philvarner: I think it would be -- I think that's one of the overall problems with these docs -- they do a good job of describing everything, but I don't they give a good picture of the dynamics of the system and how an individual pipeline would be implement and flow end-to-end. I think it's later work.

src/cirrus/docs/components/tasks/batch.rst
```diff
@@ -416,11 +416,78 @@ CPUs each job requires.
...
 *************************
 
-.. TODO
-
+Using Spot instances instead of On-Demand can result in a significant cost savings, with only a
+small effect on throughput and latency. Workflow error handling must be well-defined, since Spot
+instances have a significantly higher liklihood of being deallocated as compared with On-Demand
+instances.
+
+This is an example of a ComputeEnvironment that uses Spot::
```
stark525, line 427: extra `:`

src/cirrus/docs/components/tasks/batch.rst
```diff
@@ -416,11 +416,78 @@ CPUs each job requires.
...
+
+The AllocationStrategy of `BEST_FIT_PROGRESSIVE` indicates that Spot requests should be made
+progressively for the instance types that ECS determines will best meet the resource needs, but
+that any of these machines is acceptable. This is different than the `BEST_FIT` allocation
+strategy that will pick the best one and wait until that best one can be fulfilled.
+
+These specific `InstanceTypes` were chosen because they match the processor architecture needed
+by the Docker image that will run on them (AMD64) and they most closely "pack" with the amount
+vCPU and memory configured for the container, in this case, 1 vCPU and 9.8GB memory, with their
+1 vCPU to 8GB memory ratios.
```
stark525, line 485: If you choose an AllocationStrategy with a wide range of available instance types, larger instances will be running more jobs, and you need to ensure you have appropriate storage for these large instances. For example, if each task requires 3GB of disk space, you may be able to allocate an 8x sized instance to run 8 tasks, but would need 24GB storage. This would be a waste for smaller instance types running 1 task, where the storage is over-allocated.

src/cirrus/docs/components/tasks/index.rst
```diff
@@ -193,14 +195,89 @@ Lambda and Batch
...
 This command does the same as both of the above ``create`` command examples, so
 the listed caveats of both apply here: ensure the handler code is completed,
 and the batch configuration is updated to match the task requirements.
 
+* :doc:`Conditionally Using Batch or Lambda <../workflows/batch>`
+
+Docker Image
+---------------
+
+Unfortuantely, the same Docker image cannot be used by both Batch and Lambda, as they
```
stark525, line 211: I thought we had a way for tasks to `RunBatchAsLambda` and `RunLambdaAsBatch` but I can't seem to find it anymore.

src/cirrus/docs/components/workflows/batch.rst
```diff
@@ -1,10 +1,10 @@
 Batch tasks in workflows
```
stark525, line 1: A large section that would be great to have is Bring Your Own Container (BYOC) - "So you have a docker container, how do you modify it to work with cirrus?"

I think we've accepted that this is not going to be a first-class citizen since it varies wildly, but general guidance here would be very helpful, internally and externally.

src/cirrus/docs/components/tasks/index.rst
philvarner: @matthewhanson updated one the examples to use a Choice, and that attempted to use the same image, which didn't work. Matt, do you know of what Dan's referencing?

src/cirrus/docs/components/workflows/definitions.rst
```diff
@@ -147,6 +147,49 @@ workflows. Cirrus allows several means of building such multi-stage pipelines:
   conditions don't quite fit with the in-the-box approaches.
 
 
+Error handling
+--------------
+
```
stark525, line 152: Random thought - another thing that is confusing is the pre- and post-batch steps surrounding batch tasks in a workflow. Somewhere, we should explain why they exist, but also that a batch task may be marked as successful but fail in the post-batch step.

Caveat: I'm not sure if this is because the projects I've been on were "doing error handling wrong", often it was BYOC with a wide variety of logging and error handling strategies that we were trying to black box.

src/cirrus/docs/components/tasks/batch.rst
philvarner: If you choose an AllocationStrategy with a wide range of available instance types,
larger instances will be running more jobs, and you need to ensure you have
appropriate storage for these larger instances. Unfortunately, only one value can be
set for storage, which every EC2 instance gets, no matter how many concurrent tasks it
can support. For example, if each task requires 1vCPU and 3GB of storage and you allow
r4.xlarge (4 vCPUs) to r4.24xlarge (96 vCPUs), you must specify the the maximum amount of
storage required (96 x 3GB = 288GB), even though the r4.xlarge instances will always have
276GB of unused storage that incurs cost.

src/cirrus/docs/components/workflows/batch.rst
philvarner: this is a good point. some of this is described in the `Docker Image` section in Tasks, but I don't know if we define anywhere what the task.py should look like

src/cirrus/docs/components/tasks/index.rst
matthewhanson: There used to be a Docker image used in a generic Batch task and that could be used to run any Lambda that was deployed as a standard Lambda package (ie zip file).   The task would take in the ARN of the Lambda, fetch the deploy package code, unzip, then run the included task.py file.   It was a way to just deploy a Lambda, and not have to deploy anything else.

The utility of this isn't really there anymore.  Most importantly, we now deploy Lambdas as images to begin with, so this same image should be runnable in either Batch or Lambda. (though right now there's an issue I think with how ENTRYPOINT is set that makes these images incompatible.

Also, while the original idea of this was useful for development, in production you end up fetching the deploy package over and over again, every time you run a job.

tl;dr - now with stac-task you develop and deploy a single image, which can then be used as a Lambda or Batch task.

src/cirrus/docs/components/workflows/definitions.rst
philvarner: https://cirrus-geo.github.io/cirrus-geo/stable/cirrus/components/workflows/batch.html#pre-batch-and-post-batch ?

src/cirrus/docs/components/workflows/batch.rst
philvarner: added

stark525 (approved):

PR #249, opened 2024-02-13
SNSPublisher and SQSPublisher implementation

### LINKED ISSUE #233: SNS Publisher class
Add a basic SNS client class to abstract sending messages to SNS.
  
We may need to subclass and/or instantiate for each SNS topic managed by
cirrus.  In either case, the instances will likely live in the Workflow Event
class where a list of clients is populated and each sns-client is
instantiated with the associated ARN
 
It should support batching with some flush mechanism (context handler).

ircwaves:
As discussed in the context of the [workflow-event-topic project](https://github.com/orgs/cirrus-geo/projects/2), this extends and encapsulates the batch_handler pattern into a class.

I rolled it out to `update-state` as a test/PoC.  Feedback welcome.

/cc @philvarner @matthewhanson 

Closes #233


ircwaves:
After discussion with @jkeifer, I updated the typing in the utils module while I was working on it.

/cc @philvarner

philvarner (approved):

PR #250, opened 2024-02-13
set timestream iam perms to a valid ARN

philvarner:
1. `Condition`s removed, because these don't pass serverless validation anyway and have to be overridden
2. set the alternative Resource ARN to a syntactically valid ARN


src/cirrus/builtins/cloudformation/resources.yml
```diff
@@ -78,14 +78,12 @@ Resources:
 
   StateEventTimestreamDatabase:
     Type: AWS::Timestream::Database
-    Condition: TimestreamIsSupportedInRegion
```
stark525, line 81: Do you also want to remove the top level `Conditions` for this defined in this file?

tests/cli/fixtures/build/serverless.yml
```diff
@@ -128,7 +128,7 @@ functions:
               - Fn::GetAtt:
                   - StateEventTimestreamTable
                   - Arn
-              - Ref: AWS::NoValue
+              - arn:aws:timestream:*:*:database/nonexistent/table/nonexistent
```
stark525, line 131: Similar to above, do you want to remove the references to `TimestreamIsSupportedInRegion` in the fixture?

stark525 (approved):

tests/cli/fixtures/build/serverless.yml
philvarner: we want to keep that for the IAM policy stmts. We have the override the entire StateEventTimestreamDatabase and StateEventTimestreamDatabaseTable, so I removed it from them.

Issue #180, opened 2022-10-27
update-state needs to handle `stactask.exceptions.InvalidInput`


jkeifer:
With the introduction of stac-task, update-state is now unable to recognize this new instance of invalid input.

PR #256, opened 2024-02-15
restore InvalidInput exception

ircwaves:
This exception is still used and referenced in the docs, but worked with `cirrus-lib` (or stactask (see #180)) by using the string to triage INVALID_EXCEPTIONS.  With the purge of `cirrus-lib`, this brings that exception back to cirrus-geo.

### DIFF
--- a/src/cirrus/lib2/errors.py
+++ b/src/cirrus/lib2/errors.py
@@ -2,3 +2,9 @@ class NoUrlError(ValueError):
     """Exception class for when a payload does not have a URL."""
 
     pass
+
+
+class InvalidInput(Exception):  # noqa: N818
+    """Exception class for when processing fails due to invalid input."""
+
+    pass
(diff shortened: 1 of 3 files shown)


philvarner (approved):

philvarner:
you need to rerun pytest to update the hashes

src/cirrus/lib2/errors.py
```diff
@@ -2,3 +2,13 @@ class NoUrlError(ValueError):
     """Exception class for when a payload does not have a URL."""
 
     pass
+
+
+class InvalidInput(Exception):  # noqa: N818
+    """Exception class for when processing fails due to invalid input
+
+    Args:
+        Exception (Exception): Base class
```
ircwaves, line 11: This was direct copy-pasta from cirrus-lib, and may be wrong.
```suggestion
```

ircwaves: removed.

Issue #238, opened 2024-01-12
update lambda and lib2 to use WorkflowEventManger


ircwaves:

ircwaves:
Closed by #263

Issue #236, opened 2024-01-12
replace statedb state-change calls with Workflow Event state-change calls


ircwaves:
update all statedb state change calls to use new class (`process` lambda and
  `update-state` lambda)

ircwaves:
Closed by #263

Issue #235, opened 2024-01-12
remove timestream calls from statedb


ircwaves:
extract timestream events from statedb functions as these will be dispatched via the event manager class.

ircwaves:
Closed by #263

PR #261, opened 2024-03-05
Workflow event class

### LINKED ISSUE #234: Workflow Event class
A new python class for handling payload state change events
## needed functions
The needed functions for all state changes:
- [ ] payload started processing
- [ ] payload skipped - already processing
- [ ] payload skipped - already completed
- [ ] payload skipped - already invalid
- [ ] payload completed (success)
- [ ] payload failed
- [ ] payload invalid
- [ ] payload aborted
- [ ] payload timed out

## Notes
The WorkflowEvent class needs the following behaviors woven into the above functions:
- call statedb updates from this new class for associated state changes
- call timestream functions from this new class for associated state changes as it has been, this will only happen if timestream is enabled.
- send state-change SNS messages from this new class, for each configured client (using the new SNS client(s)), if there are any.

ircwaves:
Closes #234, #236, #238 

+ Add `WorkflowEventTopic` resource
+ Add `WorkflowEventManager` class which will be responsible for communicating state changes, and invoking StateDB methods which store state when appropriate.
+ Move `StateEnum` to `cirrus.lib2.enums` module for use across `statedb`, `eventdb`, `process_payloads`, and `events` modules.
+ Add `SfnStatus` str-enum for Step Functions execution status strings
+ Update `cirrus.core.utils.plugins` to be python 3.12 compatible.
+ Add ability to pass `boto3.Session` to `cirrus.lib2.utils.BatchHandler`.
+ Update tests for `update-state` to include `INVALID` case.
+ `*` Adds `WorkflowEventTopic` logging of malformed payloads when encountered.  

 `*` - Seeing that there were tests in the `test_process.py` which verified behavior around malformed payloads, I wondered if we should also surface this to the workflow event topic.  I'm not sure about this, as it isn't an actual `WorkflowEvent`, as the payloads in question are not properly defined.  This type of error could be left to the `ProcessDeadLetterQueue` to surface/sort-out those issues.


src/cirrus/core/utils/plugins.py
```diff
@@ -1,15 +1,16 @@
...
-
+    from importlib_metadata import EntryPoint, entry_points
 
 PLUGIN_GROUP = "cirrus.plugins"
 COMMANDS_GROUP = "cirrus.commands"
 RESOURCES_GROUP = "cirrus.resources"
 
 
-def iter_entry_points(group_name):
+def iter_entry_points(group_name: str) -> Generator[EntryPoint]:
```
jkeifer, line 13: Per the docs, I don't believe this to be a valid generator type annotation: https://docs.python.org/3/library/typing.html#typing.Generator. I think it either needs to be `Generator[EntryPoint, None, None]` or `Iterator[EntryPoint]`.

src/cirrus/core/utils/plugins.py
```diff
@@ -18,14 +19,24 @@ def sorter(ep):
...
+    else:
+        raise RuntimeError(
+            f"Unknown type returned from entry_points {type(ep_collection)}"
+        )
     eps.sort(key=sorter)
     yield from eps
 
 
-def iter_plugins():
+def iter_plugins() -> Generator[EntryPoint]:
```
jkeifer, line 37: Same annotation concern.

src/cirrus/core/utils/plugins.py
```diff
@@ -18,14 +19,24 @@ def sorter(ep):
...
     yield from iter_entry_points(PLUGIN_GROUP)
 
 
-def iter_resources():
+def iter_resources() -> Generator[EntryPoint]:
```
jkeifer, line 41: Same annotation concern.

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,88 @@
...
+import boto3
+
+from .statedb import StateDB
+from .utils import SNSPublisher
+
+logger = getLogger(__name__)
+
+
+class WorkflowEventManager:
+    """A class for managing payload state change events, inclusing storage of
```
jkeifer, line 15:
```suggestion
    """A class for managing payload state change events, including storage of
```

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,88 @@
...
+        self.announce("STARTED_PROCESSING", payload, execution_arn)
+
+    def already_processing(self, payload: Dict):
+        self.logger.warning(f"already in processing state: {payload['id']}")
+        self.announce("ALREADY_PROCESSING", payload)
+
+    def already_completed(self, payload: Dict):
+        # TODO: check with jarrett on how he sees this function used.  perhaps
+        #      raising a special exception from `StateDB.set_completed`, which forwards
+        #      over to this function?
```
jkeifer, line 67: I think you'd call this from the `else` of the `if`/`elif`/`else` at the end of `ProcessPayloads.process()`. That said, the `else` as currently written handles multiple extant states, not just `COMPLETED`, so would likely need to be broken out into separate `elif` blocks for each possible state we can handle there. We should also use the state enum instead of strings for the state matching, I think.

jkeifer: I think the states we'd handle here are at least `COMPLETED`, `INVALID`, and `PROCESSING`, but its worth considering if others would be possible.

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,88 @@
...
+        self.announce("SUCCEEDED", payload)
+
+    def invalid(self, payload: Dict):
+        self.announce("INVALID", payload)
+
+    def aborted(self, payload: Dict):
+        self.announce("ABORTED", payload)
+
+    def timed_out(self, payload: Dict):
+        self.announce("TIMED_OUT", payload)
```
jkeifer, line 88: Don't you need to set the state too?

src/cirrus/lib2/process_payload.py
```diff
@@ -22,10 +23,18 @@
...
+def get_event_manager():
+    global _event_manager
+    if _event_manager is None:
+        _event_manager = WorkflowEventManager(logger=logger)
+    return _event_manager
+
+
 def get_statedb():
     global _statedb
     if _statedb is None:
```
jkeifer, line 40: `WorkflowEventManager` has a `StateDB`, but we are also getting one here; is it worth consolidating and passing the `StateDB` instance in so we only have one?

src/cirrus/lib2/events.py
jkeifer: I see `update-state` isn't using this class yet. Is the plan to have it use this? It seems like it should.

src/cirrus/lib2/events.py
jkeifer: It looks like timestream events are still handled via `StateDB`. I was envisioning the  eventdb stuff would get pulled out of `StateDB` and handled here. Is that still the plan?

src/cirrus/core/utils/plugins.py
ircwaves: Correct you are.

src/cirrus/lib2/events.py
ircwaves: Yes.  Still the plan.  I am going to get everything to the `"use the WorkflowEventManager"` level using the `StateDB` as it is, and use #235 to address how timestream is handled. 

And thanks for the emacsfingers typo catch.

src/cirrus/lib2/events.py
ircwaves: Yeah, seems like skipping is handled the same for all payload_id states.  So, I propose this move to a `skipping` method which takes `payload`, and `state: StateEnum`, and calls `announces(f"ALREADY_{state}", payload)`.  That leaves things open to calling `skipping` on payloads in any state, which seems reasonable.

src/cirrus/lib2/process_payload.py
ircwaves: Will do.

src/cirrus/core/utils/plugins.py
ircwaves: Correct you are.

src/cirrus/lib2/events.py
ircwaves: Getting there.  But wanted to get your initial thoughts before I have this wired into all the things.  Thanks for the review!

tests/builtins/functions/test_update_state.py
```diff
@@ -44,10 +51,58 @@ def test_empty_event():
...
+    if wf_event_enabled:
+        expected_msg_count = 1
+        monkeypatch.setenv("CIRRUS_WORKFLOW_EVENT_TOPIC_ARN", workflow_event_topic)
+    else:
+        expected_msg_count = 0
+        monkeypatch.delenv("CIRRUS_WORKFLOW_EVENT_TOPIC_ARN", raising=False)
+    if "cirrus.builtins.functions.update-state.lambda_function" in sys.modules:
+        importlib.reload(
+            sys.modules["cirrus.builtins.functions.update-state.lambda_function"]
+        )
```
ircwaves, line 84: This is to deal with the state of the `wf_event_manager.event_publisher` which is initialized at import time.

tests/builtins/functions/test_update_state.py
```diff
@@ -44,10 +51,58 @@ def test_empty_event():
...
+
+    event["detail"]["status"] = sfn_state
     run_function("update-state", event)
 
     items = statedb.get_dbitems(payload_ids=[EVENT_PAYLOAD_ID])
+    assert len(items) == 1
+    assert "state_updated" in items[0]
```
ircwaves, line 91: this test is more concerned with the workflow event topic functioning, but I wanted to confirm that enabled or disabled, the state database is updated (and the state database starts out each test invocation empty, so 1 item means it did update).

tests/core/utils/test_plugins.py
```diff
@@ -0,0 +1,17 @@
+import pytest
+
+from cirrus.core.utils import plugins
+
+
+def test_iter_plugins_and_resources():
```
ircwaves, line 6: The motivation for these tests was the change in functionality with python3.12.

ircwaves (commented):
@jkeifer @philvarner @matthewhanson 

This is the first step toward integration of the workflow event manager, as laid out in [the plan](https://github.com/orgs/cirrus-geo/projects/2).  I've added some comments here as to why certain changes are made, but look forward to any feedback you all might have here.

src/cirrus/lib2/process_payload.py
```diff
@@ -247,18 +260,18 @@ def __call__(self) -> str:
             )
 
             # add execution to DynamoDB record
-            get_statedb().set_processing(self["id"], exe_response["executionArn"])
+            get_event_manager().started_processing(self, exe_response["executionArn"])
 
             return self["id"]
         except get_statedb().db.meta.client.exceptions.ConditionalCheckFailedException:
-            self.logger.warning("Already in PROCESSING state")
+            get_event_manager().already_processing(self)
```
ircwaves, line 267: add `ALREADY_PROCESSING` test

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,149 @@
...
+            "event_type": event_type,
+            "payload": payload,
+            "payload_id": payload_id,
+        }
+
+        if extra_message:
+            if not all(k not in message for k in extra_message):
+                raise ValueError(
+                    "extra_message parameters must not include "
+                    "event_type, payload, or payload_id."
```
ircwaves, line 91: make this dynamic from the message object

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,149 @@
...
+                )
```
ircwaves, line 92: The `payload_id` here is renamed to `id` in the subsequent PR #263, and the error message is made dynamic based on the keys of the `message`.

.github/workflows/python-test.yml
```diff
@@ -13,6 +13,7 @@ jobs:
           - "3.9"
           - "3.10"
           - "3.11"
+          - "3.12"
```
philvarner, line 16: also add to setup.py

CHANGELOG.md
```diff
@@ -5,6 +5,41 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- `workflow-event-topic` SNS topic, and `WorkflowEventManager` class for
```
philvarner, line 12: maybe clarify here that it's prefixed with the deployment name, since the code formatting indicates this is the exact name.

src/cirrus/builtins/cloudformation/resources.yml
```diff
@@ -14,6 +14,12 @@ Resources:
     Properties:
       TopicName: "#{AWS::StackName}-publish"
 
+  # SNS Topic for failed workflows
+  WorkflowEventTopic:
+    Type: "AWS::SNS::Topic"
+    Properties:
+      TopicName: "#{AWS::StackName}-workflow-event"
```
philvarner, line 21: everywhere else this is named as `workflow-event-topic` instead of `workflow-event`

src/cirrus/builtins/functions/process/lambda_function.py
```diff
@@ -22,8 +23,15 @@ def lambda_handler(event, context):
 
         try:
             payload = utils.extract_record(message)
-        except Exception:
-            logger.exception("Failed to extract record: %s", json.dumps(message))
+        except Exception as exc:
+            logger.exception("Failed to extract record: %s", message)
+            with WorkflowEventManager.handler(logger=logger) as wfem:
+                wfem.announce(
+                    "RECORD_EXTRACT_FAILED",
```
philvarner, line 30: maybe these should be constants in utils and documented as to their meaning?

src/cirrus/builtins/functions/update-state/lambda_function.py
```diff
@@ -46,15 +44,15 @@ class Execution:
     arn: str
     input: ProcessPayload
     output: ProcessPayload
-    status: str
+    status: SfnStatus
```
philvarner, line 47: 👍🏻

src/cirrus/builtins/functions/update-state/lambda_function.py
```diff
@@ -75,13 +73,13 @@ def from_event(cls, event: Dict[str, Any]) -> "Execution":
             status = event["detail"]["status"]
             error = None
 
-            if status == SUCCEEDED:
+            if status == SfnStatus.SUCCEEDED:
```
philvarner, line 76: if only we could use pattern matching :)

src/cirrus/builtins/functions/update-state/lambda_function.py
```diff
@@ -112,15 +110,16 @@ def workflow_completed(execution: Execution) -> None:
     # trying the sns publish, but I could see it the other
     # way too. If we have issues here we might want to consider
     # a different order/behavior (fail on error or something?).
-    statedb.set_completed(execution.input["id"], execution_arn=execution.arn)
+    wf_event_manager.succeeded(execution.input["id"], execution_arn=execution.arn)
```
philvarner, line 113: maybe this method should take `execution.input` as it's parameter to be consistent with all of the others?

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,149 @@
...
+from .enums import StateEnum
+from .statedb import StateDB
+from .utils import SNSPublisher
+
+logger = getLogger(__name__)
+
+
+class WorkflowEventManager:
+    """A class for managing payload state change events, including storage of
+    state(dynamo), activity(timestream), and notifications of decisions and/or status
```
philvarner, line 18:
```suggestion
    state (DynamoDB), activity (Timestream), and notifications of decisions and/or status
```

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def timed_out(self, payload: dict, message: str = "", execution_arn=None):
+        self.statedb.set_failed(payload["id"], message, execution_arn=execution_arn)
+        self.announce(
+            "TIMED_OUT",
+            payload,
+            extra_message={"error": message, "execution_arn": execution_arn},
+        )
+
+    def succeeded(self, payload_id: str, execution_arn: str):
```
philvarner, line 132: payload_id instead of payload, better to make it consistent I think.

src/cirrus/builtins/functions/update-state/lambda_function.py
```diff
@@ -112,15 +110,16 @@ def workflow_completed(execution: Execution) -> None:
...
     if execution.output:
+        # TODO: add test of workflow chaining
```
philvarner, line 115: are you going to do?

src/cirrus/core/utils/plugins.py
```diff
@@ -1,15 +1,16 @@
+from collections.abc import Iterator
+
 try:
-    from importlib.metadata import entry_points
+    from importlib.metadata import EntryPoint, entry_points
 except ImportError:
-    from importlib_metadata import entry_points
-
+    from importlib_metadata import EntryPoint, entry_points
```
philvarner, line 6: maybe add `# pragma: no cover`

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,149 @@
...
+    def __init__(
+        self,
+        logger: Logger = logger,
+        boto3_session: boto3.Session = None,
+        statedb: StateDB = None,
+        batch_size: int = 10,
+    ):
+        self.logger = logger
+        self._boto3_session = boto3_session if boto3_session else boto3.Session()
+        wf_event_topic_arn = os.getenv("CIRRUS_WORKFLOW_EVENT_TOPIC_ARN", None)
```
philvarner, line 30: None is redundant here

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,149 @@
...
+    ) -> AbstractContextManager["WorkflowEventManager"]:
+        wfem = cls(*args, **kwargs)
+        try:
+            yield wfem
+        finally:
+            wfem.flush()
+
+    def announce(
+        self,
+        event_type: str,
```
philvarner, line 62: I made this comment earlier, but I think the event_type could be made into an Enum instead of a string

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,149 @@
...
+        payload: dict = None,
+        payload_id: str = None,
```
philvarner, line 64: is the payload_id ever not in the payload as `id`?

tests/builtins/functions/test_update_state.py
philvarner: i hadn't seen this pattern before, good to know

src/cirrus/builtins/functions/update-state/lambda_function.py
ircwaves: so close...we could early drop py3.9 ;)

ircwaves (commented):
@philvarner -- thanks for the feedback.  I've addressed with comments.

CHANGELOG.md
```diff
@@ -5,6 +5,41 @@ All notable changes to this project will be documented in this file.
...
+
+### Added
+
+- `app-cirrus-${CIRRUS_DEPLOYMENT}-workflow-event` SNS topic, and
+  `WorkflowEventManager` class for managing workflow event actions. ([#261])
+  The actions managed by this class include:
+
+  - updating state of workflows in `StateDB`
+  - announcing interactions cirrus has with a payload to the
+    `workflow-event-topic`
```
philvarner, line 18: one more `-topic`

tests/builtins/functions/test_update_state.py
philvarner: unresolving because i commented on a note for other reviewers

tests/cli/conftest.py
```diff
@@ -31,7 +31,7 @@ def project(project_testdir):
     return Project.resolve(strict=True)
 
 
-@pytest.fixture(scope="session", autouse=True)
+@pytest.fixture(scope="module", autouse=True)
```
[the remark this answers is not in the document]
philvarner: I don't really know

philvarner (approved):

CHANGELOG.md
ircwaves: Got that, and fixed the fq-name format to the default `cirrus-<stage>-workflow-event`

CHANGELOG.md
```diff
@@ -5,6 +5,41 @@ All notable changes to this project will be documented in this file.
...
+    to their environment in both their `cirrus.yml` file:
+
+  ```yaml
+  CIRRUS_WORKFLOW_EVENT_TOPIC_ARN: !Ref WorkflowEventTopic
+  ```
+
+  and add the Topic to their `cloudformation/resources.yml` file:
+
+  ```yaml
+  # SNS Topic for failed workflows
```
matthewhanson, line 29: incorrect comment as it's not just failed workflows

matthewhanson:
This all looks good to me other than a couple comments inline.

i would suggest adding fixed attributes on the published SNS messages - State and workflow name anyway - to allow for subscription filters. But that can be a different PR since this one is large enough.

ircwaves:
@matthewhanson says:
> i would suggest adding fixed attributes on the published SNS messages - State and workflow name anyway - to allow for subscription filters. But that can be a different PR since this one is large enough.

I'll add those attributes in #263, where the WFEventType is introduced.

tests/builtins/functions/test_update_state.py
ircwaves: The reload has been made unnecessary by changes in #263

src/cirrus/builtins/cloudformation/resources.yml
```diff
@@ -14,6 +14,12 @@ Resources:
     Properties:
       TopicName: "#{AWS::StackName}-publish"
 
+  # SNS Topic for failed workflows
```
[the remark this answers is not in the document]
ircwaves: fixed in #263

CHANGELOG.md
ircwaves: fixed in #263

ircwaves:
Closing in favor of #263

PR #263, opened 2024-03-25
Migrate timestream calls to events

### LINKED ISSUE #202: statedb needs to check dynamo responses and raise errors for non-success
In troubleshooting some issues with `update-state` running at load, we found numerous examples that appeared like the db update was successful, but in fact it failed. Digging into the code, it appears we are not checking the response from dynamo to ensure writes are successful, and as a result it seems we miss errors and end up with records missing state updates.

We need to check all responses from dynamo operations and ensure success, rather than assuming.

ircwaves:
Closes #202 #234, #235, #236, #238

This supersedes #261


src/cirrus/lib2/events.py
```diff
@@ -97,14 +97,14 @@ def announce(
         message = {
             "event_type": event_type,
             "payload": payload,
-            "payload_id": payload_id,
+            "id": payload_id,
```
ircwaves, line 100: @jkeifer -- Two bits here:

1. We had discussed having `payload_id`, and I had moved to `id` but am not thinking that we should stick with `payload_id`.  An `id` element in the message feels more like it would be the message ID.  What do you think?
2. The initial proposed message format included `workflow_name` in the message.  I did not include it as it is part of the `payload_id` (assuming cirrus workflow_id convention `{collections}/workflow-{workflow_name}/{item_str}`).  Can you share any thoughts on the case for including it as a top level parameter?

CHANGELOG.md
```diff
@@ -7,10 +7,49 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
+
+  and add the Topic to their `cloudformation/resources.yml` file:
+
+  ```yaml
+  # SNS Topic for any cirrus interactions with a workflow
+  WorkflowEventTopic:
+    Type: "AWS::SNS::Topic"
+    Properties:
+    TopicName: "#{AWS::StackName}-workflow-event"
+  ```
```
jkeifer, line 34: I think you want this indented the same as a bullet's text so it will be parsed as a continuation of the bullet's content.

```suggestion
    ```yaml
    CIRRUS_WORKFLOW_EVENT_TOPIC_ARN: !Ref WorkflowEventTopic
    ```

    and add the Topic to their `cloudformation/resources.yml` file:

    ```yaml
    # SNS Topic for any cirrus interactions with a workflow
    WorkflowEventTopic:
      Type: "AWS::SNS::Topic"
      Properties:
      TopicName: "#{AWS::StackName}-workflow-event"
    ```
```

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,249 @@
...
+        )
+
+    def flush(self):
+        if self.event_publisher:
+            self.event_publisher.execute()
+
+    @classmethod
+    @contextmanager
+    def handler(
+        cls: "WorkflowEventManager", *args, **kwargs
```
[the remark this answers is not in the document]
jkeifer: That said, I see `Self` was added in 3.11...

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,249 @@
...
+            eventdb if eventdb is not None else EventDB(session=self._boto3_session)
+        )
+
+    def flush(self):
+        if self.event_publisher:
+            self.event_publisher.execute()
+
+    @classmethod
+    @contextmanager
+    def handler(
```
jkeifer, line 57: You could make `WorkflowEventManager` a context handler itself:

```
    def __enter__(self):
        return self

    def __exit__(self, et, ev, tb):
        self.flush()
```

Then to use it in a context you don't need to know/remember to call a special method, you just

```
with WorkflowEventManager() as wfm:
     ...
```

src/cirrus/builtins/functions/process/lambda_function.py
```diff
@@ -22,8 +24,15 @@ def lambda_handler(event, context):
...
-            logger.exception("Failed to extract record: %s", json.dumps(message))
+        except Exception as exc:
+            logger.exception("Failed to extract record: %s", message)
+            with WorkflowEventManager.handler(logger=logger) as wfem:
+                wfem.announce(
+                    WFEventType.RECORD_EXTRACT_FAILED,
+                    payload=message,
+                    payload_id="unknown",
+                    extra_message={"exception": str(exc)},
+                )
```
jkeifer, line 35: This appears to create a new workflow event manager just for this exception; if there are multiple exceptions you end up sending each event individually, right? Same with the other exception handler? You also have a wfem in process payload or something that is handling all other events?

Perhaps this is premature optimization, so feel free to disregard if you think so, but it _feels_ like someone makes a mistake (sending in a bunch of malformed payloads) and this becomes an problematic bottleneck. Could you instead use one wfem for all of these places you need to fire off events by creating one instance for this function, and passing it down into the `ProcessPayloads(payloads).process()` call?

Doing so seems like it would be annoying because you'd have to indent everything in this function to have it within the context of the wfem. I'd instead suggest a decorator pattern--you could add a method to the `WorkflowEventManager` class that implements a decorator that injects a keyword-only arg `wfem` into the wrapped function, and then you could use that decorator to inject `wfem` into this `lambda_handler` function.

Just a thought. We can discuss if it seems worth pursuing.

src/cirrus/lib2/eventdb.py
```diff
@@ -51,37 +44,25 @@ def __init__(
     def enabled(self) -> bool:
         return bool(self.event_db_name and self.event_table_name)
 
+    @classmethod
+    def _payload_id_to_record_data(cls, payload_id: str) -> Tuple[str, str, str]:
+        pattern = r"(?P<collections>.+)/workflow-(?P<workflow>[^/]+)/(?P<itemids>.+)"
+        expr = re.compile(pattern)
```
jkeifer, line 50: Seems like you could move this out to a constant so it doesn't have to be recompiled on every function invocation.

src/cirrus/lib2/eventdb.py
```diff
@@ -51,37 +44,25 @@ def __init__(
     def enabled(self) -> bool:
         return bool(self.event_db_name and self.event_table_name)
 
+    @classmethod
```
jkeifer, line 47: `cls` appears unused, any reason not to make this a `staticmethod` instead?

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,249 @@
...
+                                  SNS message body.
+        """
+        if self.event_publisher is None:
+            return
+
+        if payload is None:
+            if payload_id is None:
+                raise ValueError("must specify payload_id or payload")
+        elif payload_id is None:
+            payload_id = payload["id"]
```
jkeifer, line 91: Is it possible not to have an id in the payload?

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,249 @@
...
+                              exists / is  provided.
+            extra_message (dict): dictionary of additional items to be placed into the
+                                  SNS message body.
+        """
+        if self.event_publisher is None:
+            return
+
+        if payload is None:
+            if payload_id is None:
+                raise ValueError("must specify payload_id or payload")
```
jkeifer, line 89: It seems like you could collapse these conditions:

```suggestion
        if payload is None and payload_id is None:
            raise ValueError("must specify payload_id or payload")
```

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,249 @@
...
+        elif payload_id is None:
+            payload_id = payload["id"]
+
+        if payload is not None and payload["id"] != payload_id:
+            raise ValueError(
+                "payload_id and payload['id'] must match if both supplied."
+            )
```
jkeifer, line 96: I think this could be another condition on the above block, because if you are setting the `payload_id` from `payload` then you don't need to check equality.

Maybe you end up with the following?

```
        if payload is None and payload_id is None:
            raise ValueError("must specify payload_id or payload")
        elif payload is not None and payload_id is None:
            payload_id = payload["id"]
        elif payload is not None and payload["id"] != payload_id:
            raise ValueError(
                "payload_id and payload['id'] must match if both supplied."
            )
```

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,249 @@
...
+        message = {
+            "event_type": event_type,
+            "payload": payload,
```
jkeifer, line 99: What happens if the payload is larger than the max message size? Do we need to include the payload in the message? Could we provide a pointer to the payload in S3?

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,249 @@
...
+        if extra_message:
+            if not all(k not in message for k in extra_message):
+                raise ValueError(
+                    "extra_message parameters must not include: "
+                    + f"{','.join(message.keys())}."
+                )
+            message.update(extra_message)
+        self.event_publisher.add(
+            message=json.dumps(message),
+            message_attrs={
```
jkeifer, line 112: I think this was mentioned before, but adding things like the payload ID, collections, etc, could be of benefit to downstream subscribers. We should discuss what all we can add here.

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,249 @@
...
+            isotimestamp = datetime.now(timezone.utc).isoformat()
+        self.statedb.set_completed(
+            payload["id"], execution_arn=execution_arn, isotimestamp=isotimestamp
+        )
+        if execution_arn:
+            self.write_timeseries_record(
+                payload["id"], StateEnum.COMPLETED, isotimestamp, execution_arn
+            )
+        else:
+            logger.debug("set completed called with no execution ARN")
```
jkeifer, line 204: Is this possible? Can you point me to where this would be called with no ARN?

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,249 @@
...
+                payload["id"], StateEnum.FAILED, isotimestamp, execution_arn
+            )
+
+        self.announce(
+            WFEventType.FAILED,
+            payload,
+            extra_message={"error": message, "execution_arn": execution_arn},
+        )
+
+    def timed_out(
```
jkeifer, line 172: Is this not a failure with regard to the timestream events as well?

src/cirrus/lib2/process_payload.py
```diff
@@ -22,10 +25,38 @@
...
 
 
+def get_event_manager():
+    global _event_manager
+    if _event_manager is None:
+        _event_manager = WorkflowEventManager(logger=logger, statedb=get_statedb())
+    return _event_manager
+
+
+class EventManagerFlush(ContextDecorator):
```
jkeifer, line 40: This was kinda my idea with the lambda handler decorator...but it still feels like you are only using this within the bounds of a single call stack, so you could easily pass a wfem instance in.

src/cirrus/lib2/events.py
ircwaves: pointer is the way to go here.

CHANGELOG.md
ircwaves: Applied locally, for rebased reasons.

src/cirrus/lib2/events.py
ircwaves: Paired up and implemented!

src/cirrus/builtins/functions/process/lambda_function.py
ircwaves: Pursued, paired, and implemented!

src/cirrus/lib2/events.py
ircwaves: Yes.  And added.

src/cirrus/lib2/utils.py
```diff
@@ -321,6 +321,7 @@ def __init__(
         batch_size: int = 10,
         dest_name: str = "default",
         logger: logging.Logger = logger,
+        boto3_session: boto3.Session = None,
```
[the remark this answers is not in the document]
ircwaves: Do I hear a motion to punt?  @philvarner -- I think maybe some where in our discussions you had suggested adding this ability to pass in the session.  Can you clarify the potential use-case?

src/cirrus/lib2/events.py
jkeifer: I never answered this, sorry...

I think `payload_id` is better than `id`, else it seems ambiguous, like we could be meaning something like "event ID".

As to including `workflow_name`, I don't remember where that came into play, honestly. If I suggested it be part of the message itself then 🤷, but I agree it doesn't need to be part of the message.

That said, I believe it really should be part of the message attributes. It seems rather natural to expect an event listener could be concerned with only the events pertaining to a specific workflow--but as we discussed, more attributes can be added later, so I suppose it is okay to leave this as-is for now.

CHANGELOG.md
jkeifer, line 1: Can we add a `### ⚠️ Deprecated ⚠️` section and call out that the other "event" topics (`FailedTopic` and `InvalidTopic`) are deprecated and will be removed in a future release? We should tell users to migrate to the `WorkflowEventTopic` for these uses instead.

jkeifer: Well, I realized this is a problem because the current message formats for these topics is different than what you have put together for the `WorkflowEventTopic`. You're going to hate me, but can we discuss what it would take to adopt the fields from that message format (effectively defined by `StateDB.dbitem_to_item`:  https://github.com/cirrus-geo/cirrus-geo/blob/main/src/cirrus/lib2/statedb.py#L566 and  the attrs added in `update-state`: https://github.com/cirrus-geo/cirrus-geo/blob/main/src/cirrus/builtins/functions/update-state/lambda_function.py#L161)?

We don't necessarily need _all_ those fields; we can evaluate what might not be required.

While I'd like to chat about this before this PR is merged, I know there's a lot here and I understand it might be easier to make such a change in a follow-up PR instead. If we do that then we wouldn't add the deprecation until we have implemented this compatibility.

For context, my motivation is the upcoming rework project, and ensuring we can remove these topics from the deployment at that time.

jkeifer: Thinking more about this, I think this just comes down to the message attributes. As those affect how a subscriber can filter messages it receives, we have a meaningful functional difference between what can be done with the existing topics compared to the new `WorkflowEventTopic`. Looking closely at the payload differences, I think they are effectively equivalent, functionally-speaking, so I don't think we need to match the existing message format and can require subscribers to make changes to handle the new payload format.

jkeifer: Ah, though `execution_arn`, perhaps we should consider making that a url to the execution rather than just the arn.

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,305 @@
...
+        if execution_arn:
+            self._write_timeseries_record(
+                payload_id, StateEnum.ABORTED, isotimestamp, execution_arn
+            )
+        self.announce(
+            WFEventType.ABORTED,
+            payload_id,
+            payload_url,
+            extra_message={
+                "error": "none, aborted by user",
```
jkeifer, line 292: Isn't this redundant given the event type is `ABORTED`?

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,305 @@
...
+        if not isotimestamp:
+            isotimestamp = datetime.now(timezone.utc).isoformat()
+        self.statedb.set_processing(payload_id, execution_arn, isotimestamp)
+        self._write_timeseries_record(
+            payload_id,
+            state=StateEnum.PROCESSING,
+            event_time=isotimestamp,
+            execution_arn=execution_arn,
+        )
+        self.announce(
```
jkeifer, line 161: Is it worth including the timestamp in messages where we have one? Seems like that could be valuable, because the message time is not the actual event time.

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,305 @@
...
+            "payload_url": payload_url,
+            "payload_id": payload_id,
+        }
+
+        if extra_message:
+            if not all(k not in message for k in extra_message):
+                raise ValueError(
+                    "extra_message parameters must not include: "
+                    + f"{','.join(message.keys())}."
+                )
```
jkeifer, line 121: Thought for the future: I think it might be better to use something like a dataclass to more strongly type the event message format. `event_type`, `payload_url`, and `payload_id` would be required fields, and things like `error`, `execution_arn`, etc would be optional.

Then this `announce` method would only need to take a the event message type as an argument. We could also use that class as a place to centralize logic like serializing the message to json and creating the message attributes.

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,305 @@
...
+                boto3_session=self._boto3_session,
+                batch_size=batch_size,
+            )
+            if wf_event_topic_arn
+            else None
+        )
+        self.statedb = statedb if statedb is not None else StateDB.get_singleton()
+        self.eventdb = (
+            eventdb if eventdb is not None else EventDB(session=self._boto3_session)
+        )
```
jkeifer, line 50: I think we don't need the `boto3.Session()` at all on this class?

I get concerned about the `event_publisher` and `eventdb` instances because the client's are being initialized within the scope of a lambda handler. Perhaps though that we aren't seeing issues from this means that lambda bug from the past is no more?

src/cirrus/lib2/statedb.py
```diff
@@ -49,10 +93,12 @@ def __init__(
         self.table_name = table_name
         self.table = self.db.Table(table_name)
 
-        if eventdb:
-            self.eventdb = eventdb
-        else:
-            self.eventdb = EventDB(session=session)
+    @classmethod
+    def get_singleton(cls):
+        """DESCRIBE WHY"""
```
jkeifer, line 98:
```suggestion
        """When running in a lambda context we want to keep our dynamo client for
        the duration of the lambda execution environment. Using a singleton pattern
        for this case ensures we don't initialize a client for every execution."""
```

ircwaves: I knew I had more docs cleanup. Thanks for the re-review.

src/cirrus/lib2/utils.py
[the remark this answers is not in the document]
jkeifer: As we discussed the session is not required nor used by this class and should be removed.

src/cirrus/lib2/process_payload.py
```diff
@@ -354,11 +362,13 @@ def process(self, replace=False):
                 or payload["id"] in payload_ids["skipped"]
                 or payload["id"] in payload_ids["failed"]
             ):
-                logger.warning(f"Dropping duplicated payload {payload['id']}")
```
jkeifer, line 357: Was dropping this log message intentional? It still seems relevant unless I'm mistaken.

src/cirrus/builtins/functions/update-state/lambda_function.py
```diff
@@ -23,18 +25,10 @@
 SQS_CLIENT = boto3.resource("sqs")
 QUEUE = SQS_CLIENT.Queue(PROCESS_QUEUE_URL)
```
jkeifer, line 26: You don't need these anymore, right?

src/cirrus/lib2/events.py
jkeifer: 😬 The only think I have for this is a email from an AWS support interaction, which I now realize was specifically an issue with a nodejs lambda. I think I saw issues with a python lambda, but now I'm not sure anymore. Maybe we don't see issues with this now because it never affected python. Or maybe boto3 fixed the issue before the nodejs aws-sdk?

I'd say the fact that timestream client hasn't been causing an issue here means I'm just causing more problems pointing out things that aren't a problem. Except I just realized that we _have_ been using a singleton `StateDB`, which has had the reference to the timestream client--so we have gotten a singleton there by composition?

I hesitate to break that pattern here, with us specifically reinstantiating the `WorkflowEventManager` on each lambda execution. This concern also extends to the `SNSPublisher` and `SQSPublisher` classes.

Anyway, the issue was that we were seeing the error `The security token included in the request is invalid`, and it appeared to be intermittently affecting lambda executions when an execution environment was reused to process additional messages. Resolving the problem was as easy as moving our client instantiation out of the `handler`.

I am still frustrated by this issue because I could not get support to explain to me what the difference was between client inside vs outside the handler. Particularly because they said the error message usually means the temporary security credentials granted to the lambda by STS have expired, but IIRC we were seeing the issue manifest in way less than the default hour lifetime those creds get.

🤷 Not sure this helps at all.

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,353 @@
...
+
+        if match := PAYLOAD_ID_REGEX.match(self.payload_id):
+            attrs["workflow"] = {
+                "DataType": "String",
+                "StringValue": match.group("workflow"),
+            }
+            attrs["collections"] = {
+                "DataType": "String",
+                "StringValue": match.group("collections"),
+            }
```
jkeifer, line 46: Can we add just one more attribute for compatibility with the existing failed/invalid message format?

```
if self.error is not None:
    attrs["error"] = {"DataType": "String", "StringValue": error}
```

src/cirrus/lib2/events.py
```diff
@@ -0,0 +1,353 @@
...
+
+
+@dataclass
+class WorkflowEvent:
+    event_type: WFEventType
+    payload_id: str
+    isotimestamp: str
+    payload_url: Optional[str] = None
+    execution_arn: Optional[str] = None
+    error: Optional[Dict[str, Any]] = None
```
jkeifer, line 24: Is this the right type? Isn't `error` just a `str`?

src/cirrus/lib2/process_payload.py
ircwaves: I had previously toyed with moving the logging into the state-change functions, but that quickly showed itself to lead to problems.  Failed to revert this line.

src/cirrus/builtins/functions/update-state/lambda_function.py
ircwaves: Correct.

src/cirrus/lib2/events.py
ircwaves: Not the right type.  `Execution.error` in `update-state` uses a dict, but I see that it is distilled to a string before the `WorkflowEvent` is created.

src/cirrus/lib2/events.py
ircwaves: > I hesitate to break that pattern here, with us specifically reinstantiating the `WorkflowEventManager` on each lambda execution. 

Is that a motion for adopting the singleton pattern for `WorkflowEventManager`?  Let's discuss today.  

> This concern also extends to the `SNSPublisher` and `SQSPublisher` classes.

Seems like this also will need more changes then.

src/cirrus/builtins/functions/update-state/lambda_function.py
ircwaves: I addressed this before reading the back-story.  May need to discuss how that relates here.

jkeifer (approved):
🐳

PR #268, opened 2024-05-02
Initial work for v1

jkeifer:
This PR is all a starting point and we'll need to do more, but it includes:

* clean up all code no longer required with the new operational paradigm
  * `cirrus.cli` code combed over for what is relevant to the cirrus management cli and moved there, the rest removed
  * `cirrus.core` code removed
  * `cirrus.test` module removed
* reorganize remaining code
  * `cirrus.lib2` becomes `cirrus.lib`
  * `cirrus.builtins` becomes `cirrus.lambda_handlers` and flattened
* switch to `pyproject.toml` and `pip-compile` for deps
  * though I am uncertain how we should manage versions given that we _do_ want to be able to install this package into a virtual env with pip from pypi (i.e., requirements files alone ain't going to cut it)
* pull in `cirrus-mgmt` cli code and install as `cirrus` console script
  * code from https://github.com/cirrus-geo/cirrus-mgmt/pull/10 where I made the cli standalone and implemented a PoC for how to manage config from AWS resources
* docs dependency on `cirrus-docs` removed in favor of plain sphinx build and docs sources moved into `docs/`
* I also looked for additional dead code that wasn't necessarily dead just from this work and removed it.
* I pulled out all the backwards-compatibility code for deprecated features that I could find:
  * `output_options` support
  * support for the `FAILED` and `INVALID` SNS topics
  * removed `InvalidInput` exception in _this_ `cirrus.lib`
  * `legacy` API support
* I had concerns about recent changes to `BatchHandler` and the subclasses that were created, `SNSPublisher` and `SQSPublisher`. I took this opportunity to address those concerns.


jkeifer:
Tests currently fail because

1. CI needs to be updated
2. The management cli tests have not been updated per the changes to that component

I do not know if I will be able to finish the cli updates for a new config format nor fix the tests. One potential blocker is not having a new deployment to test against, which means if I do work on this I don't know if I'll derive the correct config file format. But I'll see if I can get that in.

jkeifer:
I want to question if we should also remove support for non-array `process` values, as we did in swoop... This seems like a great opportunity to do it.

jkeifer:
Tests are fixed. I am explicitly ignoring failures on the management tests right now.

ircwaves:
@jkeifer says:
> I want to question if we should also remove support for non-array `process` values, as we did in swoop... This seems like a great opportunity to do it.

Not privy to the motivations in swoop, but that is a pattern which I have seen some developers confused by, returning a list that is usually empty or 1 member.  I was implored to change it to a `-> Item | None`, with some rational that it was more idiomatic.   That said, I was never that upset by `for result in process():` in my code.  So, some sort of note that emphasizes that use-cases for `list-of-results` might be warranted.

philvarner:
Maybe lambda_handlers should be named lambda_functions?

philvarner:
wrt to deps, I think it's reasonable to have exact pinned deps, since this is the key dependency for the projects that use it

philvarner:
why "requirements files alone ain't going to cut it"?

jkeifer:
> @jkeifer says:
> 
> > I want to question if we should also remove support for non-array `process` values, as we did in swoop... This seems like a great opportunity to do it.
> 
> Not privy to the motivations in swoop, but that is a pattern which I have seen some developers confused by, returning a list that is usually empty or 1 member. I was implored to change it to a `-> Item | None`, with some rational that it was more idiomatic. That said, I was never that upset by `for result in process():` in my code. So, some sort of note that emphasizes that use-cases for `list-of-results` might be warranted.

@ircwaves I'm confused by your response and the mention of items/results. I am speaking about the process payload format, which supports an array of process definitions for workflow chaining. That is, the payload format is, roughly,

```
{
    "type": "FeatureCollection",
    "features": [Items],
    "process": ProcessDef | [ProcessDef],
}
```

What I'm proposing here is that we make `process` just `[ProcessDef]` rather than supporting the single process definition case without an array. As to why we went with that in swoop, it was to simplify the payload format by enforcing the consistency, which also reduces the test cases we have to be concerned about.

jkeifer:
> why "requirements files alone ain't going to cut it"?

Unless you know something I don't, to enable this to package to be pip-installable with the dependency pinning described in `requirements.txt` we have to have that pinning in `pyproject.toml` to have it be part of the package metadata.

philvarner:
> @philvarner
> 
> > why "requirements files alone ain't going to cut it"?
> 
> Unless you know something I don't, to enable this to package to be pip-installable with the dependency pinning described in `requirements.txt` we have to have that pinning in `pyproject.toml` to have it be part of the package metadata.

Oh, I see what you're saying now.

src/cirrus/lambda_handlers/api.py
```diff
@@ -118,20 +117,30 @@ def _results_transform(
...
 
-def hourly(r1: Dict[str, Any], *rs: Dict[str, Any]) -> List[Dict[str, Any]]:
+def hourly(
+    r1: dict[str, Any] | None,
+    *rs: dict[str, Any] | None,
+) -> list[dict[str, Any]]:
     combined_results = deepcopy(r1)
     for result in rs:
-        combined_results["Rows"].extend(result.get("Rows", []))
+        if not result:
```
philvarner, line 130: I don't think this code is right. `result` is the keys, so result should always be truthy, unless there's  something that's not apparent here, and should probably be documented.

src/cirrus/lambda_handlers/api.py
```diff
@@ -118,20 +117,30 @@ def _results_transform(
...
+            continue
+        if not combined_results:
+            combined_results = result
```
philvarner, line 133: should this break? If not, I think it would be better if there was a check above to see if combined_results is set and rs is non-empty and then pop the last one.

src/cirrus/lambda_handlers/post_batch.py
```diff
@@ -1,18 +1,19 @@
...
+from cirrus.lib.process_payload import ProcessPayload
 
-logger = get_task_logger("task.post-batch", payload=tuple())
+logger = get_task_logger("task.post-batch", payload=())
 
 BATCH_LOG_GROUP = "/aws/batch/job"
 LOG_CLIENT = boto3.client("logs")
 DEFAULT_ERROR = "UnknownError"
-ERROR_REGEX = re.compile(r"^(?:cirrus\.?lib\.errors\.)?(?:([\.\w]+):)?\s*(.*)")
+ERROR_REGEX = re.compile(r"^(?:([\.\w]+):)?\s*(.*)")
```
philvarner, line 16: why this change?

src/cirrus/lambda_handlers/api.py
jkeifer: The problem is that `eventdb.query_by_bin_and_duration` returns `dict[str, Any] | None` so we have to handle `result` potentially being `None` or we have a typing error. It looks like the root source of a potentially null value is if the eventdb is disabled, in which case we return None from the eventdb query.

src/cirrus/lambda_handlers/post_batch.py
jkeifer: This is a special case for `cirrus.lib.errors.InvalidInput` where it will strip the `cirrus.lib.errors` prefix off that error string. This maybe made sense when we were using cirrus-lib for writing tasks and that was the only `InvalidInput` exception. But now we also have `stactask.exceptions.InvalidInput`, which is what we actually want people to use, so I think the special casing here for an exception from an unmaintained lib is silly to maintain moving forward.

src/cirrus/lambda_handlers/api.py
jkeifer: I don't know, I did what I thought would preserve the existing behavior as best as I could reason about while also supporting the case where we have `None`s passed in to satisfy type checking. Maybe the whole thing deserves a bit more consideration. It might be better to throw an exception rather than returning `None` so we can avoid this case altogether.

src/cirrus/lambda_handlers/post_batch.py
jkeifer: Note that the flip side of this is `update-state` matching on the exception name to determine if a failed execution was actually invalid. That is based on the `INVALID_EXCEPTIONS` variable in `update_state.py`.

src/cirrus/lib/process_payload.py
```diff
@@ -347,7 +301,7 @@ def process(self, wfem: WorkflowEventManager, replace=False):
             _replace = replace or payload.process.get("replace", False)
 
             # check existing state for Item, if any
-            state = states.get(payload["id"], "")
+            state = states.get(payload["id"], None)
```
philvarner, line 304:
```suggestion
            state = states.get(payload["id"])
```

src/cirrus/lambda_handlers/api.py
philvarner: I think you mean it needs to check if `r1` or `rs` are None -- `result` is the values of the keys of `rs`

src/cirrus/lambda_handlers/post_batch.py
philvarner: 👍🏻

ircwaves:
> confused by your response and the mention of items/results. I am speaking about the process payload format, which supports an array of process definitions for workflow chaining. 

Ah, I see, I was thinking of the `process` function in a task. Process entry in a payload being an array of ProcessDefinitions makes sense.

src/cirrus/lambda_handlers/api.py
jkeifer: Put another way, the type annotation for `rs` of `dict[str, Any] | None` is for each argument collected into `rs` by the `*` operator, not the aggregate type of `rs`. At least that is my understanding.

jkeifer:
I rebased the changes on top of main to resolve the merge conflicts. I also renamed the `lambda_handlers` package to `lambda_functions` as suggested, and removed support for non-array process values.

jkeifer:
I talked with @gadomski and he pointed me to a solution for the dependency concern I raised. In short, we can use dynamic metadata to load dependency lists from requirements files. This means we specify the direct dependencies in `*.in` files and the `pip-compile` workflow will generate requirements files that end up populating the dependency lists in the package metadata.

I also took a super quick pass through the README and remove all obviously out-of-date content. It will need further revisions but at least isn't completely erroneous.

src/cirrus/lambda_handlers/api.py
philvarner: oh yeah, of course, I wasn't thinking about the fact that the varargs should be the type of the items in the container.

jkeifer:
@philvarner I made the changes we discussed to the API/EventDB stuff. See the most recent commit for the specific changes.

philvarner (approved):

PR #270, opened 2024-05-03
fix: passing payload_id into statedb decorator

ircwaves:


jkeifer (approved):
Your decorator descriptor implementation is quite a bit different than I've seen before, but it looks like it works and make sense.

PR #272, opened 2024-05-06
added function 'cirrus.lib2.utils.cold_start'

ircwaves:
Moves boto3 client/resource instantiation out of the invocation of the core cirrus lambdas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - `payload_id` was not passed properly to `StateDB` for logging, via the state
   change decorator. Updated the decorator to be a Descriptor, which may make
   type checking happier. ([#270])
+- added function `cirrus.lib2.utils.cold_start` to move overhead of boto
+  client/resource cache instantiation out of lambda function execution. ([#272])
 
 ### Added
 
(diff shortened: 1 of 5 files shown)


ircwaves:
@jkeifer @philvarner -- This change addresses the additional boto3 client/resource instantiation penalties which I introduced in the course of the WorkflowEventTopic project.  

Defaulting to instantiating all the clients in the core lambdas seems OK to me (the all get used) EXCEPT one: timestream.  In the case where the region doesn't support it,  @philvarner, do you know if the attempt to instantiate the `timestream-*` clients will fail in such regions?

ircwaves:
> In the case where the region doesn't support it, @philvarner, do you know if the attempt to instantiate the `timestream-*` clients will fail in such regions?

OK, in attempting to answer my own question, and recalling seeing some TimeStream messages in my CloudWatch Logs, I believe I have answered my question there -- it will be fine if I use `EventDB.enabled()` properly.  Which I have added to the PR.

jkeifer (approved):

philvarner:
> @jkeifer @philvarner -- This change addresses the additional boto3 client/resource instantiation penalties which I introduced in the course of the WorkflowEventTopic project.
> 
> Defaulting to instantiating all the clients in the core lambdas seems OK to me (the all get used) EXCEPT one: timestream. In the case where the region doesn't support it, @philvarner, do you know if the attempt to instantiate the `timestream-*` clients will fail in such regions?

I don't

PR #274, opened 2024-05-08
fix: SNSPublisher dest_name handling

ircwaves:


jkeifer (approved):

ircwaves:
gotta fix the same bits in SQS, in the same fashion.

jkeifer:
Note that I changed the merge target to the branch `release/v0`. This is going to create a merge conflict with the v1 work and we are super close to merging that, so now's the time to branch off the v0 stuff.

PR #280, opened 2024-09-24
fix #225 by preventing treatment of string as list

### LINKED ISSUE #225: Issues in `ProcessPayload.set_id()`
We definitely have a bug here: https://github.com/cirrus-geo/cirrus-geo/blob/a31362b443527d8d0df48c6c39675ef312a3c28b/src/cirrus/lib2/process_payload.py#L181. If no collection is specified on the items and `collections` is not set in the `process` block then we get a _string_ `none` and end up with an id like `n/o/n/e/...`

At the very least we need to wrap that in a tuple or something.

But generally, the whole function could use some thought. What payload fields are we still supporting? `process.collections` seems really confusing when we have `process.upload_options.collections`. What is the deal with `process.input_collections`, is that still supposed to be a thing?

ircwaves:


ircwaves:
@jkeifer -- I stopped the `push to pypi` job, as that was news to me, and didn't want to break anything.

jkeifer:
The packaging workflow only runs the push to pypi step if it is building from a tag (release). So you can run all the checks without a problem.

jkeifer (approved):
Looks good. Though you may also want to merge the CHANGELOG changes from the v0 branch into main via a different PR, as that should probably have changes pertaining to all versions.

ircwaves:
> Looks good. Though you may also want to merge the CHANGELOG changes from the v0 branch into main via a different PR, as that should probably have changes pertaining to all versions.

Regarding this, the `release/v0` branch is divergent from `main` (aka v1) work, so I implemented this change via two separate PRs, with two separate CHANGELOG entries.  

To have info about all versions, we could have an extra task for `release/v0` when releases are cut, to launch a PR to stuff that info into the `main` changelog.  Is that inline with what you're suggesting?

jkeifer:
> To have info about all versions, we could have an extra task for release/v0 when releases are cut, to launch a PR to stuff that info into the main changelog. Is that inline with what you're suggesting?

I think that idea is overkill. I'm simply pointing out we should, _at some point_, ensure the changelogs for all v0 releases end up in the changelog on main. That could be a manual process; I'm not even concerned about it in the short term, just that it eventually happens so the changelog entries aren't lost. We could, for example, make an issue to do it as a one-time thing when we determine the v0 branch is EOL.

At least for the time being I will create an issue for this so this discussion is centralized and won't be lost: #282.

Issue #188, opened 2022-11-02
New `CLAIMED` state?


jkeifer:
Processing a payload requires first "claiming" the execution by setting the state to `PROCESSING`, then actually starting the step function execution, which returns an execution ARN, and then calling `set_processing()` with that ARN to again set the state to `PROCESSING` with the addition of execution ARN.

It seems like this makes `PROCESSING` ambiguous, in the sense that it can represent both "gonna try to start processing" and "processing has started". In this sense I would propose we consider adding a `CLAIMED` state (or whatever better name makes sense here) to disambiguate between these two states.

One possible issue here is conflating payload state with execution state. Maybe this distinction is not relevant and/or useful, but `CLAIMED` is a state disconnected from an execution, whereas the rest of the states are derived specifically from an execution. In this sense, what `process` is trying to do with claiming the execution is to set a lock.

In practice I am not sure this distinction is relevant, as we still need to know if locks are stuck for things that are actually not processing or otherwise in an unexpected state. But I think recognizing the difference is potentially useful in thinking about how to best handle this situation.

Is simply adding `CLAIMED` the best solution? I once proposed using a dynamo stream to trigger a lambda to do the actual step function start, maybe that fits here? Is the problem really at the database layer and the fact that dynamo won't allow us to structure queries dynamically enough (sql in an RDBMS would trivially allow separating payloads from their individual executions with a way to query the payload's lastest state via a join, and would allow using triggers to ensure a "lock" gets cleared by an execution starting)?

jkeifer:
I think this remark is poignant:

> One possible issue here is conflating payload state with execution state. Maybe this distinction is not relevant and/or useful, but CLAIMED is a state disconnected from an execution, whereas the rest of the states are derived specifically from an execution. In this sense, what process is trying to do with claiming the execution is to set a lock.

Perhaps what we really want here is a timestamp on the dyanmo record that represent this "claim lock". If we try to again process the payload before the claim lock time has expired then the request will fail, similarly to how we currently check the state.

Using this mechanism rather than the `CLAIMED` state is we get automatic expiration--records will not get stuck in this state, and thus it behaves like the lock we want.

ircwaves:
In using the new ASL Map state to launch many jobs in sub-worklfows (thanks to [Workflow Events](https://github.com/orgs/cirrus-geo/projects/2/views/1)), hitting the case of a StateDB entry in `PROCESSING` w/o a StepFunction execution added is easily repeatable.

[...]

PR #281, opened 2024-09-25
Add claimed state to StateDB

ircwaves:
Adds `CLAIMED` state to mainline cirrus, as described in #188.


src/cirrus/lambda_functions/api.py
```diff
@@ -106,7 +106,11 @@ def _results_transform(
             "period": ts,
             "interval": interval,
             "states": [
-                {"state": state, "unique_count": state_val[0], "count": state_val[1]}
+                {
```
jkeifer, line 109: I believe we'll need a corresponding change to the dashboard to support this new state

src/cirrus/lib/events.py
```diff
@@ -153,12 +154,21 @@ def announce(self: Self, event: WorkflowEvent) -> None:
     def claim_processing(
         self: Self,
         payload_id: str,
+        arn_base: str,
```
jkeifer, line 157: Passing in the ARN base here then explicitly not using it in the return is a bit smelly to me. What if you simply took an ARN in here, and had a separate utility function that, given an ARN base, would return an ARN and execution_name? You'd call that function prior to this one.

jkeifer: You're also going to need to handle the case where the item is already in the `CLAIMED` state, e.g., `process` timed out before the step function could be started and/or the state updated to `PROCESSING`. I think this means in `ProcessPayloads.get_states()` you need to change

```
        return {c["payload_id"]: StateEnum(c["state"]) for c in self.state_items}
```

to something like 

```
        return {
            c["payload_id"]: {
                "state": StateEnum(c["state"]),
                "last_exec_arn": c["executions"][-1],
            }
            for c in self.state_items}
```

All uses of the states returned would obviously need to access the item state via the `"state"` key, but the reason for this change is to allow you to handle the `CLAIMED` state  appropriately in `ProcessPayloads.process()`. That is, if you have a record in the `CLAIMED` state you then need to pass the last execution ARN into `ProcessPayload.__call__()` so it can be reused (and you skip re-claiming). In cases where the state is not `CLAIMED` you'd generate a new ARN to be passed in to `ProcessPayload.__call__()`.

Thus, I think my suggestion to split the ARN generation from `claim_processing` is so you can actually call the suggested function in `ProcessPayloads.process()` to get a value to pass in to `ProcessPayload.__call__()`...

Maybe an even cleaner way to do this is to rewrite my suggestion a different way, perhaps:

```
        return {id: attrs for id, attrs in self.get_process_attrs(self.state_items)}

    @staticmethod
    def get_process_attrs(state_items):
        for si in state_items:
            payload_id = si["payload_id"]
            state = StateEnum(si["state"])

            if state == StateEnum.CLAIMED:
                exec_arn = si["executions"][-1]
            else:
                exec_arn = gen_exec_arn(arn_base, payload_id)

            yield payload_id, {
                "state": state,
                "exec_arn": exec_arn,
            }
```

src/cirrus/lib/process_payload.py
```diff
@@ -208,26 +207,33 @@ def __call__(self, wfem) -> str | None:
...
-                exe_response["executionArn"],
+                f"{arn}:{execution_name}",
                 payload_url=url,
             )
 
             return self["id"]
         except ClientError as e:
             if e.response["Error"]["Code"] == "ConditionalCheckFailedException":
+                # TODO: may need to address this check for claim and processing
+                #       separately now that we are specifying the execution name
```
jkeifer, line 236: You will, at least, need to handle a new error code for an already-existing execution with the given execution name. If you get that error you'll need to set the record to processing.

src/cirrus/lib/statedb.py
```diff
@@ -366,21 +382,18 @@ def set_processing(
...
-            "state_updated=:state_updated, updated=:updated, "
-            "executions = list_append(if_not_exists(executions, :empty_list), :exes)"
-        )
+        expr = "SET state_updated=:state_updated, updated=:updated"
         return self.table.update_item(
             Key=key,
             UpdateExpression=expr,
+            ConditionExpression=(
+                "begins_with(state_updated, :claim) " "and contains(executions, :exec)"
+            ),
```
jkeifer, line 391: I'm not sure you need to check for the execution ARN? If the state is `CLAIMED` then I _think_ we have an invariant here where executions must alread contain the execution ARN. Maybe no harm in checking I guess? I'm not sure.

I do think if you are performing two state updates in `ProcessPayload.__call__()` that could conditionally fail you need to ensure each is explicitly wrapped in an exception handler so you can differentiate between `ConditionalCheckFailedException` for each of them. But I think this is the comment you already have in that handler so I'll stop talking now.

src/cirrus/lambda_functions/api.py
ircwaves: cirrus-dashboard#8 ☑️

src/cirrus/lib/process_payload.py
ircwaves: @jkeifer -- My thought here is that this StepFunctions `ClientError` `Code="ExecutionAlreadyExists"` should be caught in an inner try-except, and then modify `StateDB.set_processing` to accept either `CLAIMED` or `PROCESSING` in the `ConditionExpression`.  There is some amount of race condition on getting the `set_processing` call if we don't make it idempotent.

Alternatively, we could add a specific `StateDB.already_processing` function which was used only in the case where we got the `ConditionalCheckFailedException` and we've already seen the `ExecAlreadyExists` error (last scenario in graph).

retry-after-timeout-scenarios

I think making `set_processing` permissive to being either in `CLAIMED` or `PROCESSING` is the simplest solution, and will proceed with that.  Wanted to touch base here, as pushing race-conditions around the codebase feels worth a little extra discussion.

jkeifer: My thinking on this is that in the race lost case we don't actually care about updating the state--we only want to set it to processing once, and we want that to be as early (as close to the step function start) as possible. I think implementing such behavior might necessitate blowing the exception handling apart to have nested try/except blocks for each expected exception.

Let me know if this comment doesn't give you enough direction and we can find a time to discuss further.

ircwaves: That comment works.  It is in-line with what we discussed previously.  I think emitting the WorkflowEvent `ALREADY_PROCESSING` is probably good enough in the race-lost case.

src/cirrus/lib/process_payload.py
```diff
@@ -198,7 +207,7 @@ def __call__(self, wfem) -> str | None:
         if not payload_bucket:
             raise ValueError("env var CIRRUS_PAYLOAD_BUCKET must be defined")
 
-        arn = os.getenv("CIRRUS_BASE_WORKFLOW_ARN") + self.process["workflow"]
+        state_machine_arn, _, execution_name = execution_arn.rpartition(":")
```
ircwaves, line 210: A piece that came up in debugging:  the workflow name is documented in two locations.

1. `self.process["workflow"]`
2. `re.sub(r".*/workflow-(?P<workflow>[^/][^/]*)/.*", r"\g<workflow>", payload["id"])`

I in implementing `gen_execution_arn` I had fallen back to the latter, but this line shows that the former is authoritative.

src/cirrus/lib/process_payload.py
ircwaves: Alright.  `ALREADY_PROCESSING` is wired in, as well as `ALREADY_CLAIMED`.  The latter may be of interest as a symptom of other issues.

src/cirrus/lib/events.py
ircwaves: I believe the revised deterministic execution naming is aligned with your suggestions here.

src/cirrus/lib/statedb.py
ircwaves: I _agree_ that this is redundant, and will remove.  Otherwise, this PR is ready for your more thorough review.

src/cirrus/lib/process_payload.py
```diff
@@ -277,14 +333,58 @@ def payload_ids(self) -> list[str]:
...
                 for i in self.statedb.get_dbitems(self.payload_ids)
             ]
             self.state_items = items
-        return {c["payload_id"]: StateEnum(c["state"]) for c in self.state_items}
+        response = dict(self.get_process_attrs(self.state_items))
+        response.update(
+            {
+                p["id"]: (
+                    None,
+                    ProcessPayloads.gen_execution_arn(p["id"], p.process["workflow"]),
```
jkeifer, line 348: ?

```suggestion
                    self.gen_execution_arn(p["id"], p.process["workflow"]),
```

src/cirrus/lib/process_payload.py
```diff
@@ -277,14 +333,58 @@ def payload_ids(self) -> list[str]:
...
         return [c["id"] for c in self.payloads]
 
-    def get_states(self) -> dict[str, StateEnum]:
+    def get_states_and_exec_arn(self) -> dict[str, tuple]:
         if self.state_items is None:
             items = [
                 self.statedb.dbitem_to_item(i)
                 for i in self.statedb.get_dbitems(self.payload_ids)
             ]
             self.state_items = items
```
jkeifer, line 342: Not that this is a suggestion to change anything, but it popped out at me because I recent learned about [the `cached_property` decorator](https://docs.python.org/3/library/functools.html#functools.cached_property). We could leverage it here like so:

```python
    @cached_property
    def state_items(self) -> SOME_TYPE_IM_TOO_LAZY_TO_LOOK_UP:
        return [
            self.statedb.dbitem_to_item(i)
            for i in self.statedb.get_dbitems(self.payload_ids)
        ]

    def get_states_and_exec_arn(self) -> dict[str, tuple]:
        response = dict(self.get_process_attrs(self.state_items))
        ...
```

Actually, now that I look at this more, it's kinda funny that we have `state_items` on `self` but no way to set it _without_ calling this method. That weirdness seems like something we should resolve, but I still won't ask for that to happen as part of this work unless my pointing it out makes it a compelling thing to fix.

src/cirrus/lib/process_payload.py
```diff
@@ -277,14 +333,58 @@ def payload_ids(self) -> list[str]:
...
+            yield payload_id, (state, exec_arn)
+
+    @staticmethod
+    def gen_execution_arn(payload_id, workflow) -> str:
+        """
+        Generate an execution arn for the given payload_id, using the state_item info if
+        given.
+        """
+        # is this the right UUID, or do we want to have an idempotent one
+        # that includes the executions: {payload_id}/{state_item['executions']}?
```
jkeifer, line 382: I like this idea. It makes a lot of sense, particularly given the use of uuid5 here is very specifically to enable deterministic uuids, which we don't really have with some (effectively) random timestamp.

src/cirrus/lib/process_payload.py
```diff
@@ -277,14 +333,58 @@ def payload_ids(self) -> list[str]:
...
+    def get_process_attrs(state_items):
+        for si in state_items:
+            payload_id = si["payload_id"]
+            state = StateEnum(si["state"]) if si["state"] is not None else None
+
+            # if workflow is found in CLAIMED state, then we want the last execution to
+            # determine where in the process of sfn execution the failure to proceeed to
+            # PROCESSING occurred.
+            if state == StateEnum.CLAIMED:
+                exec_arn = si["executions"][-1]
```
jkeifer, line 366: I wouldn't hold up the PR because of this one line not being tested, but it does seem like if we could whip up a test hitting this condition it would be a good one to have covered.

src/cirrus/lib/process_payload.py
```diff
@@ -277,14 +333,58 @@ def payload_ids(self) -> list[str]:
...
+        Generate an execution arn for the given payload_id, using the state_item info if
+        given.
+        """
+        # is this the right UUID, or do we want to have an idempotent one
+        # that includes the executions: {payload_id}/{state_item['executions']}?
+        execution_name = uuid.uuid5(
+            uuid.NAMESPACE_URL,
+            f"{payload_id}/{WorkflowEventManager.isotimestamp_now()}",
+        )
+        return os.environ["CIRRUS_BASE_WORKFLOW_ARN"] + workflow + f":{execution_name}"
```
jkeifer, line 387: 🤷 

```suggestion
        return f"{os.environ["CIRRUS_BASE_WORKFLOW_ARN"]}{workflow}:{execution_name}"
```

src/cirrus/lib/process_payload.py
```diff
@@ -198,55 +219,90 @@ def __call__(self, wfem) -> str | None:
...
-            wfem.claim_processing(self["id"], payload_url=url)
-
-            # invoke step function
-            self.logger.debug("Running Step Function %s", arn)
-            exe_response = get_client("stepfunctions").start_execution(
-                stateMachineArn=arn,
-                input=json.dumps(self.get_payload()),
-            )
+        # add input payload to s3
+        s3().upload_json(self, url)
```
jkeifer, line 228: This was previously inside a `try`/`except` block which triggered what appears to be the  functionality now captured in the `_fail_and_raise` function. But it looks like if this step fails for any reason (intermittent error, permissions problem?) we won't have that behavior anymore. Should this get wrapped in its own `try`/`except` with `except` calling `_fail_and_raise`?

src/cirrus/lib/process_payload.py
```diff
@@ -198,55 +219,90 @@ def __call__(self, wfem) -> str | None:
...
+            )
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "ConditionalCheckFailedException":
+                db_state = e.response.get("Item", "dbitem not found in response")
+                wfem.skipping(
+                    self["id"],
+                    state=StateEnum.PROCESSING,
+                    payload_url=url,
+                    message=(
+                        "started stepfunction, but could not set processing "
```
jkeifer, line 297: I might need to talk this over and understand how we get here exactly. But my concern is this: in the exception handler above we potentially don't start a step function, though we still try to set it to processing to communicate that it has started.

In the case that we find it is already processing, then the conditional check fails, right? That puts us into this block? Or I guess looking at the statedb code it means the state was not claimed.

In any case, are we actually ever hitting this block if we are actually skipping something? Not starting a step function because it already exists and then finding we can't update the state either is effectively a race between two duplicated payloads being processed in parallel process executions? I guess maybe when I think about it that way, then yeah, we do have a payload we are skipping, but it seems like it's not "started stepfunction, but could not set processing". Is there another case that could result in this code path?

Maybe the problem is we aren't tracking the state as to whether or not we successfully started a step function. Perhaps if we tracked that (via some flag var) we'd have an easier job of ensuring the proper message and message state?

Maybe my confusion could also be expressed with the question: is the case where a step function starts actually a "skipping" event? Maybe a controversial question, but do we even emit an event in this case: we knew we should have started a step function, that is it _was_ in the claimed state, but now it is not; can we assume whatever else won the race and changed the state means that the event we expected was already emitted?

ircwaves: > I might need to talk this over and understand how we get here exactly. But my concern is this: in the exception handler above we potentially don't start a step function, though we still try to set it to processing to communicate that it has started.

The process lambda is trying to handle two conditions:

1. the indeterminate state restart after a lambda timeout.
2. an actual race between two lambdas.

The latter condition may not actually be possible, though I'm not certain.
 
> In the case that we find it is already processing, then the conditional check fails, right? That puts us into this block? Or I guess looking at the statedb code it means the state was not claimed.

As implemented currently, this conditional failing means the payload is either not claimed or not the execution arn we expect.  The latter feels like it could only happen in the direct race condition.
 
> ... Not starting a step function because it already exists and then finding we can't update the state either is effectively a race between two duplicated payloads being processed in parallel process executions? 

Yes, I believe that to be correct.

> ... I guess maybe when I think about it that way, then yeah, we do have a payload we are skipping, but it seems like it's not "started stepfunction, but could not set processing". Is there another case that could result in this code path?
> Maybe the problem is we aren't tracking the state as to whether or not we successfully started a step function. Perhaps if we tracked that (via some flag var) we'd have an easier job of ensuring the proper message and message state?

The variable `started_sfn` is such a variable, and lets us have different messages in each case.  Only the direct race leads to one lambda starting the execution, and the other getting to update the database.  Ignoring the `different execution arn` condition for the moment, we could update the `started_sfn==True` message to say `"direct race condition detected: started stepfunction, but database already updated "`. And leave the `started_sfn==False` message as it is.

 

> Maybe my confusion could also be expressed with the question: is the case where a step function starts actually a "skipping" event? Maybe a controversial question, but do we even emit an event in this case: we knew we should have started a step function, that is it _was_ in the claimed state, but now it is not; can we assume whatever else won the race and changed the state means that the event we expected was already emitted?

Maybe this one is one we should try to walk through in person. The `wfem.skipping` function generates `ALREADY_*` event messages.  The main purpose of these is to avoid listeners from submitting a payload, and never hearing back from cirrus (deadlock).  Note that now with the no-cost `ReturnValuesOnConditionCheckFailure` parameter set, we get back the contents of the state DB, so we could inspect that to further pin down what went wrong.

src/cirrus/lib/process_payload.py
ircwaves: will do.

src/cirrus/lib/process_payload.py
ircwaves: That shouldn't be too bad to add.

ircwaves (commented):
Some thoughts to go with our discussion tomorrow.

src/cirrus/lib/process_payload.py
jkeifer: Yeah, I think you are right: if we can't claim then it seems quite likely we don't want to overwrite the payload in the bucket.

src/cirrus/lib/process_payload.py
```diff
@@ -198,55 +219,90 @@ def __call__(self, wfem) -> str | None:
...
+                    else StateEnum.PROCESSING
+                )
+
+                wfem.skipping(
+                    self["id"],
+                    state=db_state,
+                    payload_url=url,
+                    message=f"state before claim attempt was {previous_state}",
+                )
                 return None
```
ircwaves, line 257: this needs differential behaviors based on the state returned in the exception:

1. if claimed take the execution from the returned db record
2. if processing skip (renamed to `already_in`), return None

src/cirrus/lib/process_payload.py
jkeifer: Oh gosh, I totally missed `started_sfn`. 🤦 I guess that's what I get for trying to review at the end of a long day.

In any case, we've discussed, this is good. We just want to rename the `skipping` function to be something more obvious (`is_already` or whatever). That rename is tracked on a different comment so I'll resolve this one.

src/cirrus/lib/process_payload.py
jkeifer: Can we just make an issue for this problem and then we don't have to solve it now?

jkeifer (approved):
Approving so this can be merged at your leisure. But I'd recommend getting changelog entries in there as I don't know how you might want to detail that or split out into multiple entries.

src/cirrus/lib/process_payload.py
ircwaves: #304 is in there.

PR #293, opened 2024-12-02
Release prep v0.15.4

ircwaves:
### Fixed

- Pass payload, not string version of payload, when string version is larger
  than `MAX_PAYLOAD_LENGTH` is exceeded. ([#292])


trevorskaggs (approved):
Good catch, :shipit:

PR #294, opened 2024-12-06
Change update_state lambda to push items to SNS

cvangerpen:
Each item in an output Cirrus Process Payload will now be pushed to the CIRRUS_PUBLISH_TOPIC_ARN, if set. SNS message attributes are derived from the item.

We may want to consider which links in a workflow chain should be subject to publishing:
- each link (what this branch does); or
- leaf links; or
- allow the user to specify via toggle?


src/cirrus/lib/utils.py
```diff
@@ -296,6 +296,90 @@ def delete_from_queue_batch(messages: list[dict]) -> dict:
     return resp
 
 
+def build_item_sns_attributes(item: dict) -> dict:  # noqa: C901
```
cvangerpen, line 299: This was taken from the old publish builtin task:
https://github.com/cirrus-geo/cirrus-geo/blob/v0.15.4/src/cirrus/builtins/tasks/publish/lambda_function.py#L29

I made this a util function rather than a `ProcessPayload` method to a) avoid namespace pollution and b) because it operates on an `item` dict, not a payload.

tests/lib/test_utils.py
```diff
@@ -144,6 +145,14 @@ def test_delete_from_queue_bad_message(sqs, queue):
         utils.delete_from_queue(msg)
 
 
+@pytest.mark.parametrize("item", item_dir.glob("*.json"))
```
cvangerpen, line 148: Maybe this testing is overkill; I tried to follow existing test patterns in this repository, but I'm definitely open to alternative suggestions.

jkeifer (changes requested):
This all looks great. The only thing is missing tests for the lambda function itself; see https://github.com/cirrus-geo/cirrus-geo/blob/main/tests/lambda_functions/test_update_state.py. I'd at least ensure messages are sent to the topic when the payload is for a successful execution, though I would also consider implementing tests off that happy path if it were easy enough to do so.

jkeifer:
> We may want to consider which links in a workflow chain should be subject to publishing:
>
> * each link (what this branch does); or
> * leaf links; or
> * allow the user to specify via toggle?

If you consider workflow chaining to express a tree of executions, only the root node in the tree is currently executing. Thus, it only makes sense to publish items from that single execution, as you are doing here. If chaining is specified in the payload and the outputs from the current workflow are passed in to a new workflow, then that new workflow will also have zero or more output items that will be published by `update-state` for its execution.

Make sense?

src/cirrus/lib/process_payload.py
```diff
@@ -187,6 +193,16 @@ def get_payload(self) -> dict:
         url = self.upload_to_s3(self, payload_bucket)
         return {"url": url}
 
+    def to_sns_messages(self: Self) -> list[SNSMessage]:
```
jkeifer, line 196: Consider renaming to something more explicit like `items_to_sns_messages`.

jkeifer (approved):

PR #295, opened 2025-02-18
Feat/tr/parameter store config

theodorehreuter:
Update cirrus management module to utilize parameter store and hierarchical parameters to store and extract cirrus deployment component information


src/cirrus/management/deployment.py
```diff
@@ -75,20 +75,11 @@ def yield_deployments(
...
-        yield from DeploymentPointer.list(region=region, session=session)
-
-    @classmethod
-    def from_pointer(
-        cls,
-        pointer: DeploymentPointer,
-        session: boto3.Session | None = None,
-    ) -> Deployment:
-        return cls(session=session, **pointer.get_config(session=session))
+        yield from DeploymentPointer.list_deployments(region=region, session=session)
```
theodorehreuter, line 78: method previously shared a name with a built in

src/cirrus/management/deployment.py
```diff
@@ -75,20 +75,11 @@ def yield_deployments(
...
 
     @classmethod
     def from_name(cls, name: str, session: boto3.Session | None = None) -> Deployment:
-        dp = DeploymentPointer.get(name, session=session)
-        return cls.from_pointer(dp, session=session)
+        return DeploymentPointer.get_deployment(name, session=session)
```
theodorehreuter, line 82: method previously shard a name with a built in and caused mypy confusion

src/cirrus/management/deployment_pointer.py
```diff
@@ -15,150 +11,104 @@
...
-            case _ as unknown:
-                raise TypeError(f"Unsupported pointer type '{unknown}'")
+@dataclass
+class Components:
+    process_lambda: str
+    state_db_name: str
+    payload_bucket: str
+    dlq_queue_url: str | None
+    process_topic: str | None
+    process_queue_url: str | None
```
ircwaves, line 21: @jkeifer -- check me on this thought here, as I'm less familiar with the new hotness deployments.

@theodorehreuter -- This is looking promising.  When this is ready for a review, I'm happy to tag up on it.  A couple in-flight notes here.  One is that any env VARS you see within the `cirrus.lib` modules will likely need to be captured here so that they can be pushed into a local execution environment.  A few that come to mind are:

- `CIRRUS_BASE_WORKFLOW_ARN`
- `CIRRUS_DATA_BUCKET`
- `CIRRUS_WORKFLOW_EVENT_TOPIC_ARN`
- `CIRRUS_EVENT_DB_AND_TABLE`
- `CIRRUS_LOG_LEVEL`
- `AWS_REGION`
- `AWS_PROFILE` (optional)

The other is that, in the `Deployment` class, you'll see that there are functions for working with lambdas (listing and executing), which currently rely on the `self.stackname`.  I can see three roads with the maintaining of this functionality (in my preference order): 

1. TF populates `/cirrus/deployments/{deployment_name}/lambdas_functions/{lambda_name}`, and we pull from that set of SSM params, and populate `Components.lambda_functions : dict[str, str]` with the mapping of `lambda_name -> arn`.
2. We add `Components.component_prefix` (as a new `stackname`), and leave the filtering functionality in the Deployment class as it is. 
3. We don't.   I don't like this because I think having post-deployment tests of lambdas is beneficial for catching lambda configuration issues.

theodorehreuter: @ircwaves  thanks for the check in and I welcome some in flight thoughts.

I definitely appreciate that you keyed in on the `Components` model here since I'm frankly on the fence a bit about inclding it.  The original intet was to have it as a pydantic model to validate the components pulled down from the parameter store but at the moment I'm not actually packing the returned values from parameter store into a `Components` object because I wanted to leave users the freedom to use as much or little of parameter store as they wanted instead.  Because the only feedback a user would have that they have to use parameter store is if they try and use this management library and are informed they dont have alll the right parameters in parameter store. So now I'm just packing the parameters into an untyped `dict`.  But I'm not sure if thats the best approach or the returns should be more strictly modeled.

src/cirrus/management/deployment.py
```diff
@@ -75,30 +78,37 @@ def yield_deployments(
...
-        return cls.from_pointer(dp, session=session)
+        dp = DeploymentPointer.get_deployment_by_name(name, session=session)
+        if not dp:
+            raise DeploymentNotFoundError(
+                f"no deployment named '{name}' was found in the parameter store",
+            )
+        return cls.from_pointer(dp, session)
 
-    def get_lambda_functions(self):
+    def get_lambda_functions(self, session):
```
theodorehreuter, line 100: necessary to pass in session so we get right lambdas from the right aws region of interest.  Also why @pass_session was added to `manage.py`

src/cirrus/management/commands/manage.py
```diff
@@ -221,11 +221,15 @@ def process(deployment):
 @click.argument(
     "lambda-name",
 )
+@pass_session
```
theodorehreuter, line 224: need to pass so we search for lambdas in the right region if we have to specify region

src/cirrus/management/deployment_pointer.py
```diff
@@ -1,164 +1,157 @@
...
-    secret_arn: str,
-    region: str | None = None,
-    session: boto3.Session | None = None,
-) -> str:
-    sm = get_client("secretsmanager", region=region, session=session)
-    resp = sm.get_secret_value(
-        SecretId=secret_arn,
-    )
-    return resp["SecretString"]
+# core required vars.  Not exclusive.
```
theodorehreuter, line 15: basic required env vars.  More are allowable but these are what is checked and required at minimum to run management module and commands

src/cirrus/management/deployment.py
```diff
@@ -40,13 +45,11 @@ def _maybe_use_buffer(fileobj: IO):
 @dataclasses.dataclass
 class DeploymentMeta:
     name: str
-    created: str
-    updated: str
-    stackname: str
-    profile: str
+    prefix: str
```
theodorehreuter, line 48: removed unused fields like `created` and such for now

src/cirrus/management/deployment_pointer.py
```diff
@@ -1,164 +1,157 @@
...
         session: boto3.Session | None = None,
-    ) -> str:
-        return get_secret(
-            secret_arn=str(self),
-            region=self.region,
-            session=session,
-        )
-
+    ) -> list[Self]:
+        if name:  # if searching by name
```
theodorehreuter, line 54: In reference to conversations about too many parameters, now the only time we will pull everything will be if a user wants to list all all deployments. Searching for a individual deployment will only search the parameter path for that given deployment to limit returns.

src/cirrus/management/deployment_pointer.py
```diff
@@ -1,164 +1,157 @@
...
+        ssm = get_client("ssm", region=region, session=session)
+        resp = cls.get_parameters_by_path(ssm, prefix)
 
-class PointerType(StrEnum):
-    SECRET = auto()
+        parameters = resp["Parameters"]
 
+        # handle if paginated
+        while "NextToken" in resp:
+            resp = cls.get_parameters_by_path(ssm, prefix, resp["NextToken"])
```
theodorehreuter, line 63: I think this logic is right, it works for pulling down two mock deployments that result in pagination

tests/management/conftest.py
```diff
@@ -70,3 +76,36 @@ def basic_payloads(fixtures, statedb):
...
     )
+
+
+@pytest.fixture()
+def parameter_store_response():
+    with Path.open(PAYLOADS / "parameter_store_response.json") as f:
+        return json.load(f)["Parameters"]
+
+
+def mock_parameters(deployment_name: str, region: str = MOCK_REGION):
```
theodorehreuter, line 87: wanted to create more fully fleshed out mock parameter values to be able to populate with deployment name and region for slightly more testing customization and for parameter store testing I load in multiple mock deployments that I wanted to have materially different parameters.

src/cirrus/management/commands/manage.py
```diff
@@ -296,13 +300,15 @@ def _call(ctx, deployment, command, include_user_vars):
 
 
 @manage.command()
+@pass_session
 @pass_deployment
 @click.pass_context
-def list_lambdas(ctx, deployment):
+def list_lambdas(ctx, deployment: Deployment, session):
     """List lambda functions"""
+    click.echo(f"trying to list {ctx.obj}")
```
theodorehreuter, line 308: delete

tests/management/test_manage.py
```diff
@@ -21,66 +28,67 @@ def _manage(cmd):
...
+
+
+@pytest.mark.xfail()
+def test_process(deployment, manage, make_lambdas):
+    result = deployment('process {"a": "payload to test process command"}')
+    assert result.exit_code == 0
+    assert result.stdout.strip == json.dumps('{"a": "check"}')
 
 
 @pytest.mark.xfail()
```
theodorehreuter, line 94: I don't think this `get-path` command even exists and this can be deleted?

src/cirrus/management/deployment.py
```diff
@@ -40,13 +45,11 @@ def _maybe_use_buffer(fileobj: IO):
...
     environment: dict
-    user_vars: dict
-    config_version: int
+    user_vars: dict = field(default_factory=dict)
+    config_version: int = CONFIG_VERSION
+    profile: str | None = None  # aws user profile
```
jkeifer, line 52: I don't think you need any of these three.

`user_vars` was separate to allow providing key/val pairs separate from the `environment`, the latter of which was "refreshable" by pulling it down from the `process` lambda config. If both the `environment` and `user_vars` were merged into a single data structure it wouldn't be possible to know if vars had been removed from the `process` lambda env vars and should be deleted. But if we are moving to a model where additional vars can be specified by the user as parameters in parameter store then that solves the problem at the source and we only need one environment here.

`config_version` was relevant when this tool was managing a config file. Storing a config version in the file allows for future updates to easily change the file schema but maintain support for older versions. As we no longer have a config file we don't need this property anymore.

`profile` was always a bit suspicious to persist. The point with it was to make it easier to switch between deployments and not have to think about what profile is required to access each. But profile names are not necessarily consistent between users, which meant that persisting this profile name when trying to share a deployment config required users to coordinate on their AWS config profile names. I liked having this in here because it made it easier for the way I was using this tool, but it was never really that good of an idea.

With the way the tool works now profile must be specified by the user for every command (unless set in the env as `AWS_PROFILE`). If, in the future, we did persist the config (in a way that is user-specific, not shared), then we would likely want to bring this back. But as it is, we really just have a `environment`, and we are not persisting that in any way, so `profile` doesn't really make sense to have here.

src/cirrus/management/deployment.py
```diff
@@ -40,13 +45,11 @@ def _maybe_use_buffer(fileobj: IO):
 @dataclasses.dataclass
 class DeploymentMeta:
```
jkeifer, line 46: I'm not sure it makes sense to have this class at all anymore. All we are really tracking is a deployment name and a deployment environment--the prefix could easily be a var in that environment with no loss of utility. And we are not persisting this any longer, so having a class that can be dumped to/loaded from a file is not really useful.

I would lobby that this class goes away. Just add the name and environment properties to the `Deployment` class.

As an aside, the only reason this was a separate class was to leverage the `dataclass` as a container for the properties that would be persisted to the config file. By using inheritance like it is here it allowed the use of `dataclass.asdict()` to dump just the `DeploymentMeta` properties separate from anything added by `Deployment`.

src/cirrus/management/deployment_pointer.py
```diff
@@ -1,164 +1,161 @@
...
+class DeploymentEnvVars:
+    CIRRUS_BASE_WORKFLOW_ARN: str
+    CIRRUS_DATA_BUCKET: str
+    CIRRUS_EVENT_DB_AND_TABLE: str
+    CIRRUS_LOG_LEVEL: str
+    CIRRUS_PAYLOAD_BUCKET: str
+    CIRRUS_PREFIX: str
+    CIRRUS_PROCESS_QUEUE_URL: str
+    CIRRUS_STATE_DB: str
+    CIRRUS_WORKFLOW_EVENT_TOPIC_ARN: str
```
jkeifer, line 26: Why isn't this just a list of keys? It looks like it not being used as a dataclass anywhere?

```suggestion
REQUIRED_VARS = [
    CIRRUS_BASE_WORKFLOW_ARN,
    CIRRUS_DATA_BUCKET,
    CIRRUS_EVENT_DB_AND_TABLE,
    CIRRUS_LOG_LEVEL,
    CIRRUS_PAYLOAD_BUCKET,
    CIRRUS_PREFIX,
    CIRRUS_PROCESS_QUEUE_URL,
    CIRRUS_STATE_DB,
    CIRRUS_WORKFLOW_EVENT_TOPIC_ARN,
```

I would definitely question if you need all of these as well. `CIRRUS_LOG_LEVEL` certainly seems unnecessary here. What others are not used by the CLI commands? Is `CIRRUS_WORKFLOW_EVENT_TOPIC_ARN` used by anything? `CIRRUS_DATA_BUCKET`? `CIRRUS_EVENT_DB_AND_TABLE`? `CIRRUS_BASE_WORKFLOW_ARN`?

If we are not using vars then they should definitely not be required. Keep in mind that what was previously in the list of vars was literally the list of vars extracted from the `process` lambda's environment variable configuration, which was effectively all the env vars shared amongst all the cirrus lambdas, as it was pretty much defined by a top level list of vars passed into every lambda's config whether they needed them or not. And we didn't do any filtering on what we pulled down because why? But here where this list is used for validation, it should really be only what is actually required for tool operation.

src/cirrus/management/deployment_pointer.py
```diff
@@ -1,164 +1,161 @@
...
 
 
-class PointerObject(Protocol):  # pragma: no cover
-    account_id: str
-    region: str
+@dataclass
+class DeploymentPointer:
+    prefix: str
+    name: str
+    environment: dict
```
jkeifer, line 33: I feel like maybe this class has lost the plot...

The intent behind this class when I sketched it up was to serve as a _pointer_ to the deployment configuration. It seems like it is now actually the the deployment configuration. Which is really just the name and environment, as I commented on the `DeploymentMeta` class.

So, should this class be collapsed into the `Deployment` class? Is it worth keeping separate as a `DeploymentConfig` class, which then gets composed into the `Deployment` class? Should the idea of a deployment pointer be preserved and the configuration pieces of this class separated out to get this back to the state of being just a pointer?

I'm yet not sure which of these options would be better based on my short time staring at this code. What I am sure about is that this is no longer a `DeploymentPointer` as currently implemented.

jkeifer: Looking back at the previous implementation I was reminded about one of my motivations with the previous structure: decoupling the `Deployment` class from where is config was coming from and how it was being stored. In essence, we really just need to resolve the environment vars for a deployment necessary to interact with it. How we do so is not important whatsoever to the `Deployment` class.

Another problem is how to discover deployments. That was the idea behind the `DeploymentPointer` class: it _was_ tightly coupled to parameter store, but that coupling allowed `Deployment` to have no coupling to the underlying store of deployment pointers. And then `DeploymentPointer` was entirely decoupled from how to resolve said pointer into the actual deployment configuration.

This next level of indirection provided by the `Pointer` class: it allowed the parameters in parameters store to point to any resource that could store the environment keys/vals. I only implemented secrets manager as a backing store to start with, but the intent was to allow other stores such as a blob in S3, or potentially even a different tree of parameters representing the deployment's environment.

Now, pragmatically, do we really care about supporting multiple stores? I suppose maybe not. As long as we have one mechanism that we can agree on that works for the use-case the flexibility could potentially be viewed as unnecessary complication within the codebase and a maintenance burden. So I'm not necessarily lobbying to say that we need that flexibility. But I do think the previous design pattern was trying to define a set of abstraction layers to facilitate such a decoupling, and there _might_ be value in maintaining that decoupling.

src/cirrus/management/commands/deployments.py
```diff
@@ -16,4 +16,4 @@ def list_deployments(session: boto3.Session) -> None:
     List all project deployments (accessible via current AWS role)
     """
     for deployment in Deployment.yield_deployments(session=session):
-        click.echo(f"{deployment.name} ({deployment.pointer.resolve()})")
+        click.echo(f"{deployment.name} ({deployment.environment})")
```
jkeifer, line 19: I think the behavior of this command as you found it is misleading. For reference please see this command in the previous codebase prior to my moving the code and dorking around with these deployment pointers: https://github.com/cirrus-geo/cirrus-mgmt/blob/main/src/cirrus/plugins/management/commands/deployments.py#L24. Notice that just the deployment names were being listed. This is all we really want to see here, what deployments can we interact with.

I believe I had added `({deployment.pointer.resolve()})` to allow me to test the pointer was working. No other reason than that.

src/cirrus/management/deployment.py
```diff
@@ -75,30 +78,37 @@ def yield_deployments(
...
+        yield from DeploymentPointer.list_deployments(region=region, session=session)
 
     @classmethod
     def from_pointer(
         cls,
         pointer: DeploymentPointer,
         session: boto3.Session | None = None,
     ) -> Deployment:
-        return cls(session=session, **pointer.get_config(session=session))
+        return cls(session=session, **asdict(pointer))
```
jkeifer, line 89: This line is implying that the pointer is, in effect, an equivalent object to this class, aside from the addition of `session`. This to me is a pretty significant indication that we need to reconsider the abstractions provided by `Deployment` and `DeploymentPointer` as they are currently implemented, as they now appear to overlap significantly.

src/cirrus/management/deployment.py
```diff
@@ -75,30 +78,37 @@ def yield_deployments(
...
     @classmethod
     def from_name(cls, name: str, session: boto3.Session | None = None) -> Deployment:
-        dp = DeploymentPointer.get(name, session=session)
-        return cls.from_pointer(dp, session=session)
+        dp = DeploymentPointer.get_deployment_by_name(name, session=session)
+        if not dp:
+            raise DeploymentNotFoundError(
+                f"no deployment named '{name}' was found in the parameter store",
+            )
+        return cls.from_pointer(dp, session)
```
jkeifer, line 98: Raising the error here seem like a leak. Let the `DeploymentPointer` method throw that exception.

Also while not technically an error, the way your are using `DeploymentNotFoundError` is a bit off from the way its implementation appears to indicate. Just pass in the deployment name that was not found, the error message is already templated for you. I also don't think the additional context of the deployment not being found in parameter store is necessarily a good thing, it feels like a bit of a leak of the abstractions below. But that said, if you think this is valuable context then I would suggest modifying `DeploymentNotFoundError` to _not_ be templating the message and to have the message provided in the `DeploymentPointer` code that throws it, because `DeploymentPointer` _is_ tightly coupled to parameters store. You could potentially change the message in the `DeploymentNotFoundError` class itself, but that error is not coupled to parameter store in any way so I would argue it should have a generic message or none at all.

src/cirrus/management/exceptions.py
```diff
@@ -25,3 +25,9 @@ class NoExecutionsError(CirrusError):
     def __init__(self, payload_id, *args, **kwargs):
         msg = f"Payload has no executions: {payload_id}"
         super().__init__(msg, *args, **kwargs)
+
+
+class MissingParameterError(KeyError):
+    def __init__(self, parameter, *args, **kwargs):
+        msg = f"A required environment variable was not found: {parameter}"
```
jkeifer, line 32: You might want to accept a sequence of parameters names here in case multiple parameters are missing.

src/cirrus/management/deployment_pointer.py
```diff
@@ -1,164 +1,161 @@
...
+            for param_name, value, deployment_name in parameters
+            if deployment_name == name
+        }
+
+    @classmethod
+    def validate_vars(cls, parameters: list[tuple[str, str, str]]):
+        parameter_names = [name for name, _, _ in parameters]
+        for field in fields(DeploymentEnvVars):
+            if field.name not in parameter_names:
+                raise MissingParameterError(field.name)
```
jkeifer, line 161: Rather than using an iterative approach like this, I would offer the option of using sets. Instead of having a list of required fields like I suggested, make that a set of required fields. Then here, convert your parameter_names into a set, and subtract: `missing_fields = required_fields - provided_fields`.

In addition to this suggestion, I have the two questions:
* Why is `parameters` not a dict?
* Why is this a `classmethod` if `cls` is not used?

src/cirrus/management/deployment.py
theodorehreuter: roger.  Yep makes sense to delete this. Deleting

src/cirrus/management/deployment_pointer.py
```diff
@@ -1,164 +1,149 @@
...
+    "CIRRUS_PAYLOAD_BUCKET",
+    "CIRRUS_BASE_WORKFLOW_ARN",
+    "CIRRUS_PROCESS_QUEUE_URL",
+    "CIRRUS_STATE_DB",
+    "CIRRUS_PREFIX",
+]
 
 
 @dataclass()
 class Pointer:
```
theodorehreuter, line 26: collapsing stuff because we dont need secret pointer right now, but can add into case lgoic if needbe later. Just a paramstore pointer is available

src/cirrus/management/deployment.py
```diff
@@ -37,68 +41,52 @@ def _maybe_use_buffer(fileobj: IO):
...
 
     @classmethod
-    def from_name(cls, name: str, session: boto3.Session | None = None) -> Deployment:
-        dp = DeploymentPointer.get(name, session=session)
-        return cls.from_pointer(dp, session=session)
+    def from_name(cls, name: str, session: boto3.Session) -> Deployment:
+        try:
+            dp = DeploymentPointer.get_pointer(name, session=session)
+            return cls.from_pointer(dp, session=session)
+        except ClientError:
```
theodorehreuter, line 75: this gets raised if we can't find the pointer parameter in parameter store.  This seemed the appropriate place for this error because it keeps deployment_pointer agnostic if there is a pointer parameter or not, it just looks for it.  If it doesn't find it then we can't find a deployment and so then we can't build the `Deployment` class i.e. unable to get vars = unable to build deployment object for other work.

src/cirrus/management/deployment.py
```diff
@@ -206,17 +187,17 @@ def get_execution_by_payload_id(self, payload_id):
...
-    def invoke_lambda(self, event, function_name):
+    def invoke_lambda(self, event, function_name, session):
         aws_lambda = get_client("lambda", session=self.session)
-        if function_name not in self.get_lambda_functions():
+        if function_name not in self.get_lambda_functions(session):
             raise ValueError(
                 f"lambda named '{function_name}' not found in deployment '{self.name}'",
             )
-        full_name = f"{self.stackname}-{function_name}"
+        full_name = f"{self.environment['CIRRUS_PREFIX']}-{function_name}"
```
theodorehreuter, line 200: also related to issue of being unable to use lambdas via management cli

src/cirrus/management/deployment_pointer.py
```diff
@@ -1,164 +1,149 @@
...
         obj["_type"] = obj.pop("type")
+
         return cls(**obj)
 
-    def resolve(self) -> PointerObject:
+    def resolve(self):
         match self._type:
-            case PointerType.SECRET:
-                return SecretArn.from_string(self.value)
+            case "parameter_store":
```
theodorehreuter, line 39: can add other sources here if necessary later

src/cirrus/management/deployment_pointer.py
```diff
@@ -1,164 +1,149 @@
...
         session: boto3.Session | None = None,
     ):
+        """Get pointer to deployment in param store.  Pointer is one param"""
         ssm = get_client("ssm", region=region, session=session)
         return cls._from_parameter(
-            ssm.get_parameter(
-                Name=f"{deployment_prefix}{deployment_name}",
-            )["Parameter"],
-            prefix=deployment_prefix,
+            ssm.get_parameter(Name=f"{DEPLOYMENTS_PREFIX}{deployment_name}")[
```
theodorehreuter, line 117: get the single parameter from parameter store that will be used as the pointer to getting the rest of the env vars

tests/management/conftest.py
```diff
@@ -70,3 +74,64 @@ def basic_payloads(fixtures, statedb):
         ],
         statedb=statedb,
     )
+
+
+def mock_parameters(queue, payloads, statedb, workflow, deployment_name):
```
theodorehreuter, line 79: the essential parameters that have to be in parameter store to retrieve a 'deployment'.

tests/management/conftest.py
```diff
@@ -70,3 +74,64 @@ def basic_payloads(fixtures, statedb):
...
+            "",
+        ),
+        "CIRRUS_PROCESS_QUEUE_URL": queue["QueueUrl"],
+        "CIRRUS_STATE_DB": statedb.table_name,
+        "CIRRUS_PREFIX": f"fd-{deployment_name}-dev-cirrus",
+    }
+
+
+@pytest.fixture()
+def put_parameters(ssm, queue, payloads, statedb, workflow):
```
theodorehreuter, line 93: build deployment pointer and deployment env vars in parameter store for retrieval in any test that will need them

tests/management/test_manage.py
```diff
@@ -21,82 +22,73 @@ def _manage(cmd):
...
-def test_manage_refresh(deployment, mock_lambda_get_conf, lambda_env):
+def test_process(deployment, manage, make_lambdas):
+    result = deployment('process {"a": "payload to test process command"}')
+    assert result.exit_code == 0
+    assert result.stdout.strip == json.dumps('{"a": "check"}')
+
+
+@pytest.mark.xfail()
+@pytest.mark.usefixtures("_mock_lambda_get_conf")
```
theodorehreuter, line 90: can we delete this guy too? Or keep for refresh/rerun work that is next?

src/cirrus/management/deployment_pointer.py
jkeifer: I missed responding directly to this thread on the last review, and I think recent changes have effectively resolved most everything pointed out here. However, I do think @ircwaves mentioned something that deserves a bit more consideration: the `AWS_REGION` var.

As we've been talking about externally, the region feel like a particularly important thing to support, at least _at some point_. That doesn't have to be right now. But having the tool be able to fetch deployment configs from a central region then interact with deployments _in other regions automatically_ would be fantastic. Imagine for example a set of workflows that need to run close to data, but the data buckets are not in the same regions.

The idea we were discussing externally around how to load the parameters into the execution env and decoupling the management commands from the deployment concept would work well for supporting such a mode of operation.

All this said, I don't think this is worth trying to wrap up in this round of work. I created an issue tracking this idea: #299.

src/cirrus/management/commands/deployments.py
jkeifer: My explanation here was wrong--resolve just finds where the pointer points, not the contents of what is stored there. But with the simplification to supporting just parameter store as the deployment registry and config source I think displaying the pointer location isn't quite as valuable, so I'm fine with this being removed as I suggested. That said, knowing where the pointer is pointing is not necessarily a _bad_ thing to know, so we could consider maintaining that behavior, or potentially restoring it in the future. I don't have a strong opinion either way at the moment.

src/cirrus/management/deployment_pointer.py
```diff
@@ -1,164 +1,145 @@
...
+    "CIRRUS_PROCESS_QUEUE_URL",
+    "CIRRUS_STATE_DB",
+    "CIRRUS_PREFIX",
+]
 
 
 @dataclass()
 class Pointer:
-    _type: PointerType
+    _type: str
```
jkeifer, line 27: Why did you change this from an enum? `str` implies that any string value is valid, which is not true. I would stick with the `PointerType` enum because I think that is the most idiomatic way to limit this to a single valid string value for a type field (especially given that the idea of the `_type` field is to sure we have a backwards-compatible path to potentially extend the supported types in the future if we have a need to). You could also use a `Literal` string value here too. But the type `str` I think is insufficiently constrained.

theodorehreuter: Back to stringenum for stronger controls as suggested and backwards compatibility down the line.

src/cirrus/management/deployment.py
jkeifer: Hmm, well, I disagree about this being the appropriate place for this exception handling. 😅 Your reasoning sounds good, but I then interpret that reasoning to mean this should be implemented a bit differently.

The fact that a boto3 `ClientError` could be thrown is not obvious when calling `DeploymentPointer.get_pointer()`, in particular because it is not a boto3 function. doubly so that `ClientError` means what you are interpreting it to mean here, that the deployment was not found. Typically if you are calling a function from a third-party dependency and that can throw an exception you want to make sure you handle that exception at the call site to translate it into something more meaningful in that one place. As this code is written the implementation details of `DeploymentPointer.get_pointer()` are effectively leaking into this calling context.

Also, I'm not sure that handling `ClientError` indiscriminately like this is correct. `ClientError`, in my experience, is thrown for _any_ error the client encounters, not all of which would indicate the deployment doesn't exist (such as hitting a rate limit, network issues, etc.). Often proper handling of a `ClientError` requires further inspection of the error. [Here's an example](https://github.com/cirrus-geo/cirrus-geo/blob/main/src/cirrus/lib/process_payload.py#L244) from a different part of the cirrus codebase showing how only two specific `ClientErrors` could be handled in that context, and anything else has to be re-raised.

src/cirrus/management/deployment_pointer.py
jkeifer:
```suggestion
            case PointerType.ParamStore:
```

src/cirrus/management/deployment_pointer.py
```diff
@@ -97,68 +36,115 @@ class Pointer:
...
 
 
+class ParamStoreDeployment:
+    def __init__(self, deployment_key: str):
+        self.prefix = deployment_key
+
+    def fetch(self, session=boto3.Session) -> dict[str, str]:
+        """Get all parameters for specific deployment"""
+
+        parameters = DeploymentPointer._get_parameters(self.prefix, session)
```
jkeifer, line 57: Calling a method on a different class in this way is a particularly strong code smell. If you have logic that you want to share between two classes like this then that logic does not belong to either of them, and it should be factored out into a third shared object/source.

See my other related comment on the `DeploymentPointer` class.

src/cirrus/management/deployment_pointer.py
```diff
@@ -97,68 +36,115 @@ class Pointer:
...
+        return {
+            param["Name"].split(self.prefix)[1]: param["Value"] for param in parameters
```
jkeifer, line 59: If you use my generator rewrite of `get_parameters_by_path()` below, then you can make this like:

```suggestion
            param["Name"].removeprefix(self.prefix): param["Value"]
            for param in get_parameters_by_path(self.prefix, session)
```

Doing so avoids the intermediate allocation of the entire list. Also note the use of the string method `removeprefix` as opposed to `split`.

src/cirrus/management/deployment_pointer.py
```diff
@@ -97,68 +36,115 @@ class Pointer:
...
+        session: boto3.Session,
+    ) -> list[str]:
+        """Retrieve and list names of available deployments in parameter
+        store"""
+        parameters = cls._get_parameters(
+            DEPLOYMENTS_PREFIX,
+            session=session,
+        )
+
+        return [param["Name"].split(DEPLOYMENTS_PREFIX)[1] for param in parameters]
```
jkeifer, line 136: With the generator version of `get_parameters_by_path` I am pretty sure this can be written as:

```suggestion
    ) -> Iterator[str]:
        """Retrieve and list names of available deployments in parameter
        store"""
        yield from (
            param["Name"].removeprefix(DEPLOYMENTS_PREFIX)
            for param in get_parameters_by_path(DEPLOYMENTS_PREFIX, session=session)
        )
```

Doing so convert it to a generator as well, so call sites might also need a slight update.

src/cirrus/management/deployment_pointer.py
```diff
@@ -97,68 +36,115 @@ class Pointer:
...
+
+    def validate_vars(self, environment: dict[str, str]):
+        missing = [field for field in REQUIRED_VARS if field not in environment]
```
jkeifer, line 139: Minor nitpick, but again, you could do this as a set operation, which is likely to be more optimized under the covers. With our scale of the number of "fields" (I'm not sure what a field is in this context but whatever...) an efficiency difference won't be a concern here--I just keep saying this because if you are not up on sets then I would suggest you look into them.

src/cirrus/management/deployment.py
theodorehreuter: class was deleted per asyn convo

src/cirrus/management/deployment_pointer.py
theodorehreuter: Removed to only required env vars for CLI toosl

src/cirrus/management/deployment_pointer.py
theodorehreuter: applied

src/cirrus/management/deployment_pointer.py
theodorehreuter: ah yeah familiar with sets, your earlier comment slipped past me, I was using them in the original implementation that ironically had excess iteration despite using sets, where I was constructing the deployment names.    And yeah calling them field is off base too good point.  These are meant to be env vars, calling env vars `fields` I agree is poor naming.  I like the simpler approach cause while this is a list comprehension list comprehensions I find are generally less readable and `missing = required - found` is even simpler.

jkeifer: Okay, cool, I realize I made an erroneous assumption here, sorry about that. Again, in this specific context switching to a set operation is a bit of a nitpick, but I do agree even if performance isn't a concern the readability of the set version is a bit higher.

tests/conftest.py
```diff
@@ -170,3 +186,28 @@ def _environment() -> Iterator[None]:
     finally:
         os.environ.clear()
         os.environ = current_env  # noqa: B003
+
+
+@pytest.fixture()
+def execute_state_function(stepfunctions, workflow, put_parameters):
```
jkeifer, line 192: Minor nit: I would choose to name this `execute_state_machine` or `execute_step_function`. AWS Step Functions leverages the concept of a "state machine" to execute a multi-step workflow, thus "state function" sounds a little funny. 😁

src/cirrus/management/deployment.py
theodorehreuter: outdated

tests/management/conftest.py
```diff
@@ -25,6 +27,8 @@ def timestream_write_client():
 
 LAMBDA_ENV_VARS = {"var": "value"}
 
+PAYLOADS = Path(__file__).parent / "fixtures"
+
```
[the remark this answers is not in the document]
theodorehreuter: yes.  Deleted.

src/cirrus/management/deployment_pointer.py
```diff
@@ -97,68 +36,115 @@ class Pointer:
...
+        ssm = get_client("ssm", region=session.region_name, session=session)
+        resp = cls._get_parameters_by_path(ssm, prefix)
+        parameters = resp["Parameters"]
+
+        # handle pagination
+        while "NextToken" in resp:
+            resp = cls._get_parameters_by_path(ssm, prefix, resp["NextToken"])
+            parameters = parameters + resp["Parameters"]
+
+        return parameters
```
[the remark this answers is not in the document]
theodorehreuter: had to tweak next token assignment to `next_token = resp.get('NextToken')` so that it wouldn't throw a `KeyError` if a next token isn't there.  Param Store just doesn't have a `NextToken` in the response object at all if theres no paginaton, so was throwing a key error with the bracket get.   Now if we're at the end of pagination NextToken will get set to None and break the while loop.  Can't initially set NextToken to `None` because `.get_parameters_by_path` doesn't allow `NextToken=None` it would seem.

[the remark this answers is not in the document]
jkeifer: Then do `next_token = resp.get("NextToken", "")`.

src/cirrus/management/deployment_pointer.py
theodorehreuter: I think I'm starting to get a feel for `yield`, I really like that you can avoid an intermediate list assignment like this, the efficency/memory improvement is obvious.

tests/conftest.py
theodorehreuter: lmao.  Yep.

src/cirrus/management/exceptions.py
```diff
@@ -25,3 +25,9 @@ class NoExecutionsError(CirrusError):
     def __init__(self, payload_id, *args, **kwargs):
         msg = f"Payload has no executions: {payload_id}"
         super().__init__(msg, *args, **kwargs)
+
+
+class MissingParameterError(KeyError):
+    def __init__(self, parameters, *args, **kwargs):
+        msg = f"A required environment variable(s) was not found: {parameters}"
+        super().__init__(msg, *args, **kwargs)
```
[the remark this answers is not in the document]
theodorehreuter: what's the purpose of doing this extra logic here in the error vs doing it as currently?  A single missing parameter or multiple missing parameters are still going to come in as just a single string by doing string concatenation on missing parameters in `deployment_pointer.py`, so theres no risk of additional args getting passed in which this fix seems to be looking to handle.  

Plus as I think about this, do we really even need to have this be a subclass of `KeyError`? Cause. its not really a key error, this isn't coming from a failed l lookup, its just coming from a failure to get something from parameter store that is required by the CLI tool.  If anything, I think just having it raised from the base `CirrusError` makes more sense, since this is really an error unique to cirrus itself - we can't find things that are important for a cirrus deployment (and might in fact indicate an improper set up of cirrus).   If we were directly trying to access a parameter and it was missing that might make more sense to raise it from `KeyError` but that's not the usecase.

src/cirrus/management/deployment_pointer.py
theodorehreuter: Yeah I'm really liking `yield` helping avoid those intermediate assignments.

src/cirrus/management/deployment_pointer.py
theodorehreuter: Excellent point.  When I had eliminated the pointer class, it was more sensible to keep it in the `DeploymentPointer` class but as you point out, its use is now in two different classes.   Pulled out into a stand alone utility function refactord per your rec with 'yield'

src/cirrus/management/commands/deployments.py
theodorehreuter: I think for simplicity's sake I prefer to just be listing out the names themselves, it would be easy to add back in later, and printing out 'param_store' next to every deployment for now feels like UI clutter for a user.

src/cirrus/management/deployment.py
theodorehreuter: outdated.  Archiving.

src/cirrus/management/exceptions.py
theodorehreuter: yeah its just coming in as a concatenated string now to keep it simpler than handling multiple inputs, just a single string input of what parameters weren't found.

src/cirrus/management/deployment.py
theodorehreuter: Interesting.  I get what you mean in regards to handling error from a 3rd party API call at the site of the API that makes sense instead of letting it float up and be handled elsewhere, possibly making it harder to track.   That's a helpful example to link to.

src/cirrus/management/deployment_pointer.py
theodorehreuter: +1 for readability always.  Went with sets.

src/cirrus/management/deployment_pointer.py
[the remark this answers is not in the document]
theodorehreuter: I know we're getting pretty deep in the weeds here but what's the argument for that a superior approach vs
`
next_token = resp.get("NextToken")
if not next_token:
     break
`

Is the argument that an affirmative check against an actual string is superior over a negative check against None?   Or that here we're keeping the value of next_token always a valid string, either `""` or the token string.

src/cirrus/management/deployment.py
theodorehreuter: Moved error handling into `deployment_pointer.py` to wrap the API call that could/should raise the error if we can't find a deployment.

ircwaves (approved):
LGTM.

src/cirrus/management/commands/manage.py
```diff
@@ -221,11 +246,15 @@ def process(deployment):
 @click.argument(
     "lambda-name",
 )
+@pass_session
 @pass_deployment
-def invoke_lambda(deployment, lambda_name):
+def invoke_lambda(deployment: Deployment, session: boto3.session, lambda_name: str):
```
ircwaves, line 251: That is a module, `Session` is the type.
```suggestion
def invoke_lambda(deployment: Deployment, session: boto3.session.Session, lambda_name: str):
```

src/cirrus/management/commands/manage.py
```diff
@@ -122,7 +132,7 @@ def run_workflow(deployment, timeout, raw, poll_interval):
         timeout=timeout,
         poll_interval=poll_interval,
     )
-    click.echo(json.dump(output, sys.stdout, indent=4 if not raw else None))
+    json.dump(output, sys.stdout, indent=4 if not raw else None)
```
ircwaves, line 135: I could be wrong here, but I think `click.echo` was here to get a newline into the console.  I'd suggest keeping it.  `json.dump` doesn't append a newline. So click.echo is a little hacky, but effective at keeping the terminal looking nice, no matter what environment you are in:

```python
>>> json.dump({"as": "you were"}, sys.stdout)
{"as": "you were"}>>>
>>>
>>> click.echo(json.dump({"as": "you were"}, sys.stdout))
{"as": "you were"}
>>>
```

src/cirrus/management/commands/manage.py
```diff
@@ -296,13 +325,14 @@ def _call(ctx, deployment, command, include_user_vars):
 
 
 @manage.command()
+@pass_session
 @pass_deployment
 @click.pass_context
-def list_lambdas(ctx, deployment):
+def list_lambdas(ctx, deployment: Deployment, session: boto3.session):
```
ircwaves, line 331:
```suggestion
def list_lambdas(ctx, deployment: Deployment, session: boto3.session.Session):
```

ircwaves (changes requested):
Couple small changes....but looks good!

src/cirrus/management/commands/manage.py
theodorehreuter: ahh.  Yeah mypy wasn't happy after typing the inputs because `json.dump` doesn't technically return anything.   Could just use the type ignore since I agree having the output cleaner is much nicer.  Mypy didnt mind before it was typed but once I typed everything mypy had stuff to get upset about.

src/cirrus/management/commands/manage.py
```diff
@@ -112,7 +116,12 @@ def show(deployment):
 )
 @raw_option
 @pass_deployment
-def run_workflow(deployment, timeout, raw, poll_interval):
+def run_workflow(
+    deployment: Deployment,
+    timeout: int,
+    raw: str | None,
```
jkeifer, line 122: This type is weird to me. Isn't this a `bool` with a default of `False`?

src/cirrus/management/commands/manage.py
```diff
@@ -122,7 +131,7 @@ def run_workflow(deployment, timeout, raw, poll_interval):
         timeout=timeout,
         poll_interval=poll_interval,
     )
-    click.echo(json.dump(output, sys.stdout, indent=4 if not raw else None))
+    click.echo(json.dump(output, sys.stdout, indent=4 if not raw else None))  # type: ignore
```
jkeifer, line 134: I saw there was some other chatter about this line--I think this is a bit of nit, but just throwing out the option to do

```suggestion
    click.echo(json.dumps(output, indent=(4 if not raw else None)))
```

for consistency with other CLI output where we are using `dumps` not `dump`. I'm not sure the possible performance difference here is meaningful one way or the other. I don't think we are really concerned here about memory usage and thus don't want the intermediate string allocation--which probably happens regardless. If that is truly a concern then we'd probably need to be reevaluating the rest of the code in that regard. As a result I'd lobby for this suggested change.

src/cirrus/management/commands/manage.py
```diff
@@ -131,7 +140,7 @@ def run_workflow(deployment, timeout, raw, poll_interval):
 )
 @raw_option
 @pass_deployment
-def get_payload(deployment, payload_id, raw):
+def get_payload(deployment: Deployment, payload_id: str, raw: str | None):
```
jkeifer, line 143: Again, isn't `raw: bool = False`?

src/cirrus/management/commands/manage.py
```diff
@@ -160,7 +169,12 @@ def download(output_fileobj):
 @execution_arn
 @raw_option
 @pass_deployment
-def get_execution(deployment, arn, payload_id, raw):
+def get_execution(
+    deployment: Deployment,
+    arn: str | None,
+    payload_id: str | None,
+    raw: str | None,
```
jkeifer, line 176: Same, isn't `raw: bool = False`?

src/cirrus/management/commands/manage.py
```diff
@@ -174,7 +188,12 @@ def get_execution(deployment, arn, payload_id, raw):
 @execution_arn
 @raw_option
 @pass_deployment
-def get_execution_input(deployment, arn, payload_id, raw):
+def get_execution_input(
+    deployment: Deployment,
+    arn: str | None,
+    payload_id: str | None,
+    raw: str | None,
```
jkeifer, line 195: Same, isn't `raw: bool = False`?

src/cirrus/management/commands/manage.py
```diff
@@ -188,7 +207,12 @@ def get_execution_input(deployment, arn, payload_id, raw):
 @execution_arn
 @raw_option
 @pass_deployment
-def get_execution_output(deployment, arn, payload_id, raw):
+def get_execution_output(
+    deployment: Deployment,
+    arn: str | None,
+    payload_id: str | None,
+    raw,
```
jkeifer, line 214: And no type here? Isn't `raw: bool = False`?

src/cirrus/management/commands/manage.py
```diff
@@ -235,10 +263,10 @@ def invoke_lambda(deployment, lambda_name):
 @include_user_vars
 @pass_deployment
 def template_payload(
-    deployment,
-    additional_variables,
-    silence_templating_errors,
-    include_user_vars,
+    deployment: Deployment,
+    additional_variables: dict[str, str] | None,
```
jkeifer, line 267: Why the `| None`? If the problem is that `additional_variables` doesn't get a value by default then we should be setting one, not allowing it to be `None`. In other words, an empty dict is a perfectly valid representation of not having any additional variables, and would be preferable to the possibility of having a null value.

Said one last way: seeing a container data type (list, dict, set, etc.) optionally be `None` is a pretty strong code smell. Recognize it as a problem upstream from here--the solution is not to add `| None` to the typing but to make sure whatever is providing this is providing a reasonable value. So the `additional_variables` function from `utils/click.py` should have a default value of `{}` specified.

----

An exception to this smell is unfortunately a problem with the way python allows specifying optional function parameters, in that if you want to have a an optional keyword parameter by giving it a default value then you most often do need to use `None` as a sentinel value which can be detected in the function to properly initialize an empty container value:

```python
def example(optional_param: dict[str, Any] | None = None):
    if optional_param is None:
        optional_param = {}
```

Because you often really don't want to do the traditional footgun:

```python
def example(optional_param: dict[str, Any] = {}):
```

src/cirrus/management/commands/payload.py
```diff
@@ -42,7 +44,12 @@ def get_id():
 @payload.command()
 @additional_variables
 @silence_templating_errors
-def template(additional_variables, silence_templating_errors):
+def template(
+    additional_variables: dict[str, str],
```
jkeifer, line 48: Ah, so here there is no `| None`. Does this mean that we don't need to set the default value to `{}`? It is already coming through as `{}`?

src/cirrus/management/commands/deployments.py
jkeifer: Just a clarification: I wasn't saying that the output could say that the source is always parameters store, but what the prefix is of the deployment's parameters. For example, say we have two deployments:

```
/cirrus/deployments/deployment_a: parameters under /cirrus/parameters/deployment_a/
/cirrus/deployments/deployment_b: parameters under /some/other/prefix/for/deployment_b/
```

Then what I was saying is that the output here could be like:

```
deployment_a (/cirrus/parameters/deployment_a/)
deployment_b (/some/other/prefix/for/deployment_b/)
```

But choose your own adventure here. 🤷

src/cirrus/management/deployment_pointer.py
```diff
@@ -97,68 +60,92 @@ class Pointer:
...
+            return cls._from_parameter(parameter, name=deployment_name)
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "ParameterNotFound":
+                raise DeploymentNotFoundError(deployment_name) from e
+            raise
 
     @classmethod
-    def list(
+    def list_deployments(
         cls,
```
jkeifer, line 128: `cls` is unused. I would suggest that means this should be a `@staticmethod`.

src/cirrus/management/deployment_pointer.py
```diff
@@ -97,68 +60,92 @@ class Pointer:
...
+        yield from (
+            param["Name"].removeprefix(DEPLOYMENTS_PREFIX)
+            for param in get_parameters_by_path(DEPLOYMENTS_PREFIX, session=session)
+        )
+
+    def validate_vars(self, environment: dict[str, str]):
+        missing = REQUIRED_VARS - set(environment.keys())
+        if missing:
+            raise MissingParameterError(", ".join(missing))
+        return environment
```
jkeifer, line 143: `self` is unused. I would suggest that means this should be a `@staticmethod`.

src/cirrus/management/deployment_pointer.py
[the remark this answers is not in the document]
jkeifer: The point is to keep from having a null. It sounds like where `next_token` is being used that a `None` is not allowed, so it should never have that type.

As for the suggestion `next_token = resp.get("NextToken") if not next_token: break`--is this valid python? Also it is checking the value of `next_token` _before_ calling `resp.get("NextToken")`. Python is a little backwards with its "ternary" operator syntax; if we rewrite this using a more standard C-like syntax I think the problem becomes immediately clear, maybe?

```
next_token = next_token ? break : resp.get("NextToken")
```

Is that what you meant to propose?

Again, is this even valid? `break` is a statement (it doesn't return a value), not an expression (which does), so I don't think it is syntactically valid to use it in assignment like this, in any language I know. Well, okay, `break` is an expression in rust and _can_ return a value depending on the context, but that's, well, even rust is inconsistent about where that "feature" can be used. 😅 It also returns the value from the loop, which means even there you can't use `break` in an assignment statement like this...

src/cirrus/management/exceptions.py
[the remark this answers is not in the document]
jkeifer: `KeyError`? I guess I misunderstood how you were using this, I just stuck with `KeyError` because you had that as the base class. Understanding more about how you are using this, I agree with the change to `CirrusError`. You could inherit from both `KeyError` and `CirrusError` if you needed to, but again I agree that this is not a `KeyError`.

As for my suggestion to move the join into the error: it's because you are already using it for templating one or more var names into the error message. It seems like a cleaner and more reusable API to accept one or more var names as args and do the join in the function. Said differently, if the point of overriding the constructor of this error is to make encapsulate the logic required to get a consistent error message, then encapsulate _all_ of that logic.

src/cirrus/management/commands/payload.py
theodorehreuter: So I would think that for this `template` command, the expected behavior for no additional variables is that you would just get back the input payload as is.   And that is accomplished with it as it currently is in the `merge_vars1`. 

Is adding a `{}` as default for `additional_variables` more for readability, that we have a default value and that default for typing should also be noted in the function definition even though that default is coming from elsewhere?

src/cirrus/management/commands/manage.py
theodorehreuter: Ah still confused by command line.  I thought that since it goes in as a string in the command line since I'm typing it in as a string, I thought everything comes in as a string.  Looking at the definition of `@raw_option` there isn't a `default` defined which `click.option` allows, so I got thrown.  Though I could see how logically an optional input at the command is True if present and False if not.   I thought an absence would be None.

src/cirrus/management/commands/manage.py
theodorehreuter: tweaked to `json.dumps` per Jarrets suggestion further down

src/cirrus/management/commands/manage.py
jkeifer: `raw_option` is defined as

```python
def raw_option(func):
    return click.option(
        "-r",
        "--raw",
        is_flag=True,
        help="Do not pretty-format the response",
    )(func)
```

The `is_flag` option makes this a boolean option, meaning that the absence of the flag will result in a value of `False` by implicitly setting `default=False`.

But go read the click docs. Yes, inherently everything that comes into the program within `argv` is a string, but click supports types. Don't just assume types, make sure you know what they are. In some case, obviously here, mypy isn't going to be of much help validating this. I think the point is more that the signature of `run_workflow` is yours to define, and if the click options are providing the wrong types you'd want to flag that. As opposed to typing the function to the click options, that's the wrong way around.

src/cirrus/management/commands/payload.py
theodorehreuter: It looks to already be coming through as `{}` from https://github.com/cirrus-geo/cirrus-geo/blob/45ebfa598635c14d41259e5c82674675a6862820/src/cirrus/management/utils/click.py#L152 where the default is two steps away.

For something like this where the default is getting injected by is abstracted away, should we put a default here?  I would argue yes to make the function more obvious.  There is a default that is coming through from `merge_vars1` but that's not immediately obvious looking at this function, and indicating it does have a default value seems beneficial.

src/cirrus/management/commands/manage.py
theodorehreuter: It looks like `additional_variables` has effective defaults through the `merge_vars1` and `merge_vars2` but that's not immediately obvious given the multiple abstractions away from where these inputs are actually being handled in `template`.   Ill add a default value to the click option/argument itself so its more clear that a default value does exists, and its clear what that default is.

src/cirrus/management/commands/payload.py
theodorehreuter: Good catch, they do already come through as `{}` from `additional_variables`.  Per your other suggestion I'll make sure defaults are explicitly set in `additional_variables`

src/cirrus/management/exceptions.py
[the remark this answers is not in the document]
theodorehreuter: Gotcha, applied this tweak.

jkeifer (approved):

PR #301, opened 2025-03-13
Updating Cirrus docs for CLI parameter store changes

theodorehreuter:
Updating the cirrus documentation after reworking the cirrus cli management tool to use parameter store for holding deployment environment variables.  Currently in parallel with https://github.com/cirrus-geo/cirrus-geo/pull/295 where the CLI changes are being made.

### DIFF
--- a/docs/src/cirrus/cli/00_intro.rst
+++ b/docs/src/cirrus/cli/00_intro.rst
@@ -6,19 +6,17 @@ Introduction
 Cirrus itself is `STAC`_-based geospatial processing pipeline platform deployed
 on AWS and is a constellation of associated pieces of AWS infrastructure.  The
 pieces of a Cirrus deployment like the lambdas, SQS, S3 buckets and other can be
-managed and interacted with using common AWS management tools, either from the
-AWS GUI or AWS CLI.  However the nature of a Cirrus deployment can make direct
+managed and interacted with using common AWS management tools like the AWS GUI
+or AWS CLI.  However the nature of a Cirrus deployment can make direct
 management via AWS tools difficult to do things like batch reruns, manage
-deployments, individual invokations, and more.  There can be multiple cirrus
-deployments, and Cirrus infrastructure like SQS queues, S3 buckets, and lambdas
-may co-exist in an AWS account with non Cirrus resources further complicating
-this management.
+deployments, individual lambda invocations, and more.  There can be multiple
+cirrus deployments and Cirrus infrastructure may co-exist in an AWS account with non Cirrus resources further complicating this management.
 
 What Is CLIrrus?
 ----------------
-To ease the management of a cirrus deployment, the cirrus-geo library has a
+To ease the management of a cirrus deployment the cirrus-geo library has a
 built in command line tool (CLIrrus) designed to streamline the developer
-experience when using and managing a Cirrus deployment.
+experience when using and interacting with a Cirrus deployment.
 
 CLIrrus is a python CLI tool built using the `click`_ library with commmands
 specific to Cirrus.  Example use cases would be to list all available cirrus
--- a/docs/src/cirrus/cli/00_intro.rst
+++ b/docs/src/cirrus/cli/00_intro.rst
@@ -42,7 +42,7 @@ CLIrrus Quick Start
 -------------------
 If you understand the broader background and structure of Cirrus, cirrus
 deployments and the cirrus CLI tool and you want to jump right into using the
-CLI tool here is what an example workflow might look like:
+CLI tool here is what an example workflow might look like.  These are one way of accomplishing these steps, personal and organizatonal requirements may differ.
 
 1. Install requirements
 
@@ -50,7 +50,7 @@ CLI tool here is what an example workflow might look like:
 
     python3.12 -m venv .venv
     source .venv/bin/activate
-    pip install cirrus-geo>=1.0.0
+    pip install 'cirrus-geo[cli]>=1.0.0'
 
 2. Authenticate (if necessary)
 
--- a/docs/src/cirrus/cli/00_intro.rst
+++ b/docs/src/cirrus/cli/00_intro.rst
@@ -50,9 +50,9 @@ CLI tool here is what an example workflow might look like:
 
     python3.12 -m venv .venv
     source .venv/bin/activate
-    pip install -r requirements-dev.txt -r requirements-cli.txt -r requirements.txt
+    pip install cirrus-geo>=1.0.0
 
-2. Authenticate
+2. Authenticate (if necessary)
 
 .. code-block:: bash
 
@@ -63,14 +63,6 @@ CLI tool here is what an example workflow might look like:
 .. code-block:: bash
 
     cirrus mgmt deployment-name-here list-lambdas
-
-If your cirrus resources are in a different region from the account sso region
-you can use a region flag
-
-.. code-block:: bash
-
-    cirrus --region us-west-2 mgmt deployment-name-here list-lambdas
-
 
 And thats it!  If you would like more in depth explanations on how to use the
 tool, please proceed.
--- a/docs/src/cirrus/cli/00_intro.rst
+++ b/docs/src/cirrus/cli/00_intro.rst
@@ -1,0 +1,82 @@
+CLIrrus - The Cirrus CLI management Tool
+========================================
+
+Introduction
+------------
+Cirrus itself is `STAC`_-based geospatial processing pipeline platform deployed
+on AWS and is a constellation of associated pieces of AWS infrastructure.  The
+pieces of a Cirrus deployment like the lambdas, SQS, S3 buckets and other can be
+managed and interacted with using common AWS management tools, either from the
+AWS GUI or AWS CLI.  However the nature of a Cirrus deployment can make direct
+management via AWS tools difficult to do things like batch reruns, manage
+deployments, individual invokations, and more.  There can be multiple cirrus
+deployments, and Cirrus infrastructure like SQS queues, S3 buckets, and lambdas
+may co-exist in an AWS account with non Cirrus resources further complicating
+this management.
+
+What Is CLIrrus?
+----------------
+To ease the management of a cirrus deployment, the cirrus-geo library has a
+built in command line tool (CLIrrus) designed to streamline the developer
+experience when using and managing a Cirrus deployment.
+
+CLIrrus is a python CLI tool built using the `click`_ library with commmands
+specific to Cirrus.  Example use cases would be to list all available cirrus
+deployments, list lambdas in a given deployment, and even directly trigger a
+workflow with an input Payload by sending the payload directly to the SQS queue
+monitored by the Process lambda, and more.
+
+What if I'm using an older version of Cirrus?
+---------------------------------------------
+
+You can use the `older CLI tool`_ which is a standalone module and can be
+installed via pip.
+
+How To Use CLIrrus?
+-------------------
+A few simple steps are all it takes to set up CLIrrus for use.
+
+1. Install
+2. Authenticatie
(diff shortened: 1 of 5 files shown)


docs/src/cirrus/cli/00_intro.rst
```diff
@@ -0,0 +1,80 @@
...
+    aws sso login --profile your-config-profile-here
+
+3. Use CLIrrus
+
+.. code-block:: bash
+
+    cirrus mgmt deployment-name-here list-lambdas
+
+If your cirrus resources are in a different region from the account sso region
+you can use a region flag
```
ircwaves, line 68: Is there another reason for this distinction?  Every time I've setup an SSO profile, it's included:

`sso_region` - where the sso auth needs to happen

and

`region` - the region where I'm going to be working.

docs/src/cirrus/cli/01_install.rst
```diff
@@ -0,0 +1,16 @@
...
+
+For superior environment management we recommend you create a specific virtual environment.  This can be accomplised using your preferred environment management tools.  One way might be as follows:
+
+.. code-block:: bash
+
+    python3.12 -m venv .venv
+
+    source .venv/bin/activate
+
+    pip install -r requirements-dev.txt -r requirements-cli.txt -r requirements.txt
```
ircwaves, line 13: Same as previous comment+suggestion in installation command.

docs/src/cirrus/cli/00_intro.rst
```diff
@@ -0,0 +1,80 @@
...
+
+.. code-block:: bash
+
+    cirrus --region us-west-2 mgmt deployment-name-here list-lambdas
+
```
ircwaves, line 73: Feels like this would also need `--profile your-config-profile-here`, or to suggest that you `export AWS_PROFILE=your-config-profile-here`.

docs/src/cirrus/cli/02_auth.rst
```diff
@@ -0,0 +1,35 @@
...
+names of your choice.
+
+Required key value pairs are:
+
+- sso_session: A user assigned name given to the session
+- sso_account_id: AWS account ID to use to authenticate
+- sso_role_name: A valid AWS role assigned to the account ID
+- region: AWS Region to SSO into
+
+After you have built your config file you can login with
```
ircwaves, line 26: I'd say this section should be replaced with pointer to [AWS documentation](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html), and a nudge to follow the best practices of the users organization.  The configuration here is different from what I've been using with aws cli 2.8.6.  I consistently end up with these keys in my sso profile configs (~/.aws/config):

- sso_start_url - url the cli will hit to start the authorization flow.
- sso_region - region where the sso accounts are configured
- sso_account_id - aws account ID 
- sso_role_name - valid role to assume via sso
- region - region you will be using in conjunction with this configuration
- output - format of cli output you would like

ircwaves (changes requested):
As this has mostly gotten into the auth and packaging side of things, I'm going to break this review in two.  Part two coming up after second breakfast.

docs/src/cirrus/cli/00_intro.rst
```diff
@@ -0,0 +1,80 @@
...
+1. Install requirements
+
+.. code-block:: bash
+
+    python3.12 -m venv .venv
+    source .venv/bin/activate
+    pip install -r requirements-dev.txt -r requirements-cli.txt -r requirements.txt
+
+2. Authenticate
```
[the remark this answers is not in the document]
theodorehreuter: ah excellent point, I agree generalizing it would be better.  I'm inclined to leave this in as an example of ONE WAY you could do since I think just saying "make sure youve configured aws creds to the same account" is a bit too non specific.

docs/src/cirrus/cli/00_intro.rst
```diff
@@ -0,0 +1,80 @@
...
+deployments and the cirrus CLI tool and you want to jump right into using the
+CLI tool here is what an example workflow might look like:
+
+1. Install requirements
+
+.. code-block:: bash
+
+    python3.12 -m venv .venv
+    source .venv/bin/activate
+    pip install -r requirements-dev.txt -r requirements-cli.txt -r requirements.txt
```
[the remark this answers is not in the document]
theodorehreuter: Good point.  I think having 4 different requirement files is also silly frankly, and would vastly prefer poetry and a single `pyproject.toml` and lock file, but I know that we're not doing that right now.

docs/src/cirrus/cli/00_intro.rst
theodorehreuter: That's me regurgitating my approach.  Been focused on moving foward stuff just got something working with SSO but as you rightly point out you can set it up differently.   In the docs I do want to specify how to use regions like this for someone who might not be super familiar with AWS stuff and CLIs.  Case in point me haha, I was stumped for a hot minute early on when my region that got picked up was the account region but then needed to interact with stuff in a different region.  You are right that the config could be more finely tuned to avoid this example but I would like to have an example of using a meta option.

docs/src/cirrus/cli/00_intro.rst
theodorehreuter: it actually not need that.  The AWS Profile right now feels kinda extraneous, I've got all the commands working locally without needed to include profile in any of them, and I don't include profile in my aws config file either.

docs/src/cirrus/cli/02_auth.rst
theodorehreuter: +1 for following 'best practices of your organization'

@ircwaves  queston about the first line though where you say 

> sso_start_url - url the cli will hit to start the authorization flow.

Is this essential though?  My currnet config doesn't have it and I get automatically redirected to the client specific login portal, I suspect pulled from the account ID that is redirecting me there.   

I'm thinking these docs should have the MINIMUM necessary requirements to get someone up and running and I don't think `sso_start_url` or `output` or even `region` are required though I'd agree including `region would be preferable to avoid entering it for every command.  Thoughts on doing MINIMUM required fields vs setting folks up with best practice?

docs/src/cirrus/cli/00_intro.rst
theodorehreuter: After thinking about it just eliminating the entire bit about region meta argument.  Keeping it as short and simple as possible to make the quick-start truly quick.

docs/src/cirrus/cli/01_install.rst
theodorehreuter: See my response to your response above re: this installation command

docs/src/cirrus/cli/00_intro.rst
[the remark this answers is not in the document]
ircwaves: Happy to huddle/follow-up on this, but AFAIK `'cirrus-geo[cli]>=1.0.0'` is the union of `requirements.txt` and `requirements-cli.txt`.  So, it doesn't include `dev` requirements, but those aren't needed for running the CLI.  We could get there with `cirrus-geo[cli,dev]>=1.0.0`, but those aren't in the `pyproject.toml` yet.  And, I agree the requirements management needs some love, and that should be separate.

As for the lockfile, I did not enjoy `poetry` when I tried it.  I do however enjoy the `uv` experience, which provides locking.

docs/src/cirrus/cli/04_commands.rst
```diff
@@ -0,0 +1,181 @@
...
+        cirrus mgmt name-dev get-execution --payload-id sar/workflow-test/example-01_2024-10-31-06-05-10
+
+        cirrus mgmt name-dev get-execution --arn arn:aws:states:us-west-2:000000000011:execution:fd-name-dev-cirrus-project:c123456789-b19292-999
+
+- *get-execution-input:*
+    Get a workflow execution's input payload using ARN or payload-id
+
+    .. code-block:: bash
+
+        cirrus mgmt kodiak get-execution-input --arn arn:aws:states:us-west-2:000000000011:execution:fd-name-dev-cirrus-project:c123456789-b19292-999
```
ircwaves, line 60:
```suggestion
        cirrus mgmt name-dev get-execution-input --arn arn:aws:states:us-west-2:000000000011:execution:fd-name-dev-cirrus-project:c123456789-b19292-999
```

docs/src/cirrus/cli/04_commands.rst
```diff
@@ -0,0 +1,181 @@
...
+    Invoke lambda with event (from stdin) and specifying by name which lambda to invoke
+
+    .. code-block:: bash
+
+        cirrus mgmt name-dev invoke-lambda process
+        {
+            "type": "FeatureCollection",
+            "process": [{"workflow": "example"}],
+            "features": []
+        }
```
ircwaves, line 97: This doesn't seem like it is showing a bash command reading from stdin.  Maybe this:
```suggestion
        echo '{"type": "FeatureCollection", 
               "process": [{"workflow": "example"}],
               "features": []}' \
        | cirrus mgmt name-dev invoke-lambda process
```

ircwaves: same idea applies for the other stdin commands.  Might even be more succinct to use

```
cat payload.json | cirrus mgmt name-dev invoke-lambda process
```

theodorehreuter: I like this for being more succinct.  I'm still a newb with command line and json stuff, so I didn't know about piping and using cat to feed in output till you showed me this example.  Thanks.

docs/src/cirrus/cli/04_commands.rst
theodorehreuter: fixed

docs/src/cirrus/cli/00_intro.rst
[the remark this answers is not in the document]
theodorehreuter: ah I hadn't encountered that `library[extra]` syntax before.  Sounds good.

docs/src/cirrus/cli/04_commands.rst
jkeifer: I hate to be _that guy_, but this is an [unnecessary use of `cat`](https://porkmail.org/era/unix/award) and I'm gonna have to call it out. All of the `cat`s I saw in the examples are unnecessary, just do `<payload.json cirrus mgmt name-dev invoke-lambda process` or `cirrus mgmt name-dev invoke-lambda process <payload.json` (they are equivalent, pick which you prefer, I like the former).

Of course I you are one of those people that feels like you _have_ to use `cat` here for some reason I can't understand then please proceed... 😅

theodorehreuter: that's a good link.  We aren't actually concatenating anything, just need to pipe the payload in.

docs/src/cirrus/cli/00_intro.rst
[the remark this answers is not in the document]
jkeifer: The dev requirements _not_ being in the extras is by design: I feel pretty strongly that dev requirements should not be part of the package's install-time API because dev packages _should not_ be required for any runtime behavior--if that is incorrect then said dev dependencies are not dev dependencies and should be moved to the normal list (or a different extra, if more appropriate).

As for having requirements files instead of putting the deps into the pyproject.toml: this is the only way I've found that allows for a pip-compile-type workflow to pin application dependencies strictly. If this were a lib with lose dependency pinning I would do that in pyproject.toml. But as an application we want strict pinning managed via pip-compile (or an equivalent), and this is the way to get that version pinning into the package metadata.

In short, I have strong opinions here about why and how dependencies are managed, which I believe I can defend quite well. I'm happy to have a discussion to convince you all that this is indeed the way. 😁

jkeifer (approved):

PR #303, opened 2025-03-24
Use CLI IAM role for sessions

theodorehreuter:
Terraform now creates (optionally) a new CLI IAM role in the parameter store.  If this specific parameter is present, then it is included in the download of other related deployment parameters.  This IAM ARN can be either passed in by a user at the manage command point, or retrieved from the parameter store

This PR uses this CLI IAM role to retrieve new credentials associated with that role and update the credentials in the session object with these new more narrowly targeted permissions.

A new `assume_role` function is execute at the time of creation of the deployment object.  if the CLI parameter was extracted the session object will be updated, otherwise the original session will be returned.

Added a few tests unrelated to new sessions to get more test coverage of commands


src/cirrus/management/commands/manage.py
```diff
@@ -82,13 +82,23 @@ def wrapper(*args, **kwargs):
...
+)
 @pass_session
 @click.pass_context
-def manage(ctx, session: Session, deployment: str, profile: str | None = None):
+def manage(
+    ctx,
+    session: Session,
+    deployment: str,
+    profile: str | None = None,
+    iam_arn: str | None = None,
```
theodorehreuter, line 96: custom passing in arn role by user at command line level if desired.

tests/management/test_manage.py
```diff
@@ -71,50 +119,17 @@ def test_list_lambas(deployment, make_lambdas, put_parameters):
...
 def test_process(deployment, manage, make_lambdas):
     result = deployment('process {"a": "payload to test process command"}')
     assert result.exit_code == 0
     assert result.stdout.strip == json.dumps('{"a": "check"}')
 
 
-@pytest.mark.xfail()
-@pytest.mark.usefixtures("_mock_lambda_get_conf")
-def test_manage_refresh(deployment, lambda_env):
-    result = deployment("refresh")
```
theodorehreuter, line 92: We dont have any 'refresh' command in the command line right now so removing this

tests/management/test_manage.py
```diff
@@ -71,50 +119,17 @@ def test_list_lambas(deployment, make_lambdas, put_parameters):
...
+def test_manage_show_deployment(deployment, put_parameters):
+    result = deployment("show")
     assert result.exit_code == 0
-    new = json.loads(deployment("show").stdout)
-    assert new["environment"] == lambda_env
-
-
-@pytest.mark.xfail()
-@pytest.mark.usefixtures("_environment")
-def test_manage_get_execution_by_payload_id(
```
theodorehreuter, line 100: this paradigm of executing process directly and calling `deployment.get_execution_by_payload_id`seems incorrect for testing a cli command, so I replaced it with a test similar the others where the test makes a command line call of `get-execution --payload-id`

tests/management/conftest.py
```diff
@@ -132,3 +136,51 @@ def lambda_handler(event, context):
         Description="mock process lambda for unit testing",
     )
     return lambdas
+
+
+@pytest.fixture()
+def create_records(s3, put_parameters, statedb, payloads, st_func_execution_arn):
```
theodorehreuter, line 142: This is pretty large and an argument could be made for trying to make it smaller, but the `payload-id` based relation between records in statedb and the s3 bucket is necessary for the `get-payload` test and later work on the `rerun` where we filter on the statedb records and then use those records to get payloads.  That relation I feel is a strong argument for this rather large fixture.

tests/management/conftest.py
```diff
@@ -132,3 +136,51 @@ def lambda_handler(event, context):
...
+            "sar-test-panda/workflow-test/failed-1",
+        ],
+    }
+
+    # add to mock statedb first then to mock payload bucket
+    # repeat use of st_func_execution_arn for get execution tests
+    # set processing to executio arn is also in mock statedb for other tests
+    for index, id in enumerate(payload_ids["processing"]):
+        (
+            statedb.set_processing(
```
theodorehreuter, line 166: need to set two successive states because only `set_processing` actually inserts an `execution_arn` into the statedb and then that `execution_arn` is used to get executions for some commands/testing those commands.

tests/management/test_manage.py
ircwaves: This is to test the deployment class method which gets the latest execution, as it also is leveraging the `executions[-1]` magic.  I'd leave the test, or rewrite the method to dry-out getting the most recent execution (but maybe that'd be better as a follow up issue).

ircwaves (approved):
Overall looks good.  Couple small changes, and then good by me.

PR #305, opened 2025-04-01
Adding bulk get_payloads command

theodorehreuter:
This PR adds a new `get_payloads` command.  Per ticket:

"With the new deployment model via terraform we are dropping the last remaining core feeder, feed-rerun. Some operators/projects still depend on that feeder, so we need to ensure we have a compatible alternative available. As a result, we need to add a command to the cirrus-mgmt cli and document how to use it to replace uses feed-rerun.

Specifically, we need a command to query state database for multiple records that supports the following parameters:
    Collections
    Workflow name
    State
    since timestamp
    Limit
    Error begins with string

Additional parameters could also be added to the above if we see specific value, but the above is the minimal list we need to support for compatibility with feed-rerun.

Once we have that command in place, we can document how to use its output to trigger reprocessing of the payloads therein.

 Idea is to use xargs or an equivalent to iterate through the input payloads, fetch them, then pipe the fetched payloads to the process command for enqueuing in the pipeline

 We need to ensure we set "replace": true in the payload so it isn’t just skipped."

Per Jarrett, we want each record to come out as a new line delimited JSON to facilitate pipeling and the piping operations as specified in the ticket.


src/cirrus/management/commands/manage.py
```diff
@@ -1,79 +1,38 @@
...
+    query_filters,
+    raw_option,
+)
 
 logger = logging.getLogger(__name__)
 
 pass_deployment = click.make_pass_decorator(Deployment)
 
 
-def _get_execution(
```
theodorehreuter, line 27: moving a number of utility functions out of the command file to clean it up

src/cirrus/management/commands/manage.py
```diff
@@ -334,3 +283,48 @@ def list_lambdas(ctx, deployment: Deployment, session: Session):
...
+    state: str | None,
+    since: str | None,
+    error_prefix: str | None,
+    limit: int = 100,
+):
+    """
+    Query multiple records from state DB using filter options and pipe to
+    stdout as new line json to facilitate stream processig/xargs piping
+    """
+    os.environ.update(deployment.environment)
```
theodorehreuter, line 303: bringing parameter store vars stored in deployment object into local env for client calls

src/cirrus/management/commands/manage.py
```diff
@@ -334,3 +283,48 @@ def list_lambdas(ctx, deployment: Deployment, session: Session):
...
+                download_payload(deployment, item["payload_id"], b)
+                b.seek(0)
+                payload = json.load(b)
+
+            payload["replace"] = True
+
+            # echo sends to stdout as NDJSON for piping into xargs
+            click.echo(json.dumps(payload, default=str))
+        except botocore.exceptions.ClientError as e:
+            logger.error(e)
```
theodorehreuter, line 330: could make this more specific based on likely error coming back from `download_payload`

tests/management/conftest.py
```diff
@@ -132,3 +134,54 @@ def lambda_handler(event, context):
         Description="mock process lambda for unit testing",
     )
     return lambdas
+
+
+@pytest.fixture()
+def create_records(
```
theodorehreuter, line 140: open to suggestions to making this setup fixture smaller.  However we do need stuff in both mock statedb and mock payload bucket for this test (and other command tests added in IAM PR #303)

src/cirrus/management/utils/manage.py
```diff
@@ -0,0 +1,103 @@
...
+    )(func)
+    func = optgroup.option(
+        "--since",
+        help="Time filter of how far back to search records",
+    )(func)
+    func = optgroup.option(
+        "--limit",
+        help="limit the options returned ",
+        default=100,
+        type=click.IntRange(1, 50000),
```
theodorehreuter, line 44: set high because hearing from users like Ian it sounds like there can be 10k's of failed runs this command is trying to help rerun in bulk

tests/management/test_manage.py
```diff
@@ -127,3 +128,27 @@ def test_manage_get_execution_by_payload_id(
...
+):
+    assert result.exit_code == 0
+    output = result.stdout.strip().split("\n")
+    assert len(create_records[state]) == len(output)
+    for payload in output:
+        assert json.loads(payload)["payload_id"] in create_records[state]
+
+
+def test_get_records(deployment, create_records, statedb):
+    result = deployment(
```
theodorehreuter, line 146: only testing for two filters instead of more combinations because transformations into queries are handled in statedb.py, only trying to test command functionality and that it comes back as NDJSON

src/cirrus/management/utils/manage.py
```diff
@@ -0,0 +1,103 @@
...
+        "--state",
+        help="Execution state to filter on",
+    )(func)
+    func = optgroup.option(
+        "--since",
+        help="Time filter of how far back to search records",
+    )(func)
+    func = optgroup.option(
+        "--limit",
+        help="limit the options returned ",
```
pjhartzell, line 42: This wording is a little confusing to me. Does this mean an option to limit the number of records returned?

pjhartzell (changes requested):
Left some comments and questions in the implementation code. I can look at the tests once we clear out these comments.

CHANGELOG.md
```diff
@@ -11,12 +11,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
...
 - `CLAIMED` state to StateDB, along with associated `ALREADY_CLAIMED`
   WorkflowEvent. ([#281])
+- `get_records` command to `manage` command group to facilitate bulk retrieval of payloads to be piped into other commands for bulk reruns of failed jobs ([#305])
 
 ### Changed
 
 - StepFunctions executions now use a deterministic UUID5, derived from the
   payload ID and execution history ([#281])
 - Upgrade `moto` to enable testing of `process` lambda race conditions ([#281])
+- Moved some utility functions for management commands into their own utils file ([#305])
```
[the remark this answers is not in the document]
theodorehreuter: shortening, and adding markdown linter.

src/cirrus/management/commands/manage.py
```diff
@@ -334,3 +283,48 @@ def list_lambdas(ctx, deployment: Deployment, session: Session):
...
+    )
+
+    for item in items["items"]:
+        try:
+            with BytesIO() as b:
+                download_payload(deployment, item["payload_id"], b)
+                b.seek(0)
+                payload = json.load(b)
+
+            payload["replace"] = True
```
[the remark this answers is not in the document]
theodorehreuter: Documentation does not make this clear but you are correct, I checked with Ian.  Fixed.

src/cirrus/management/utils/manage.py
theodorehreuter: Yeah.  Tweaked that to say `payload` since that's what is getting returned.  This eventually turns into the LIMIT parameter when we make the client query to dynamo db.  Looking at the docs it seems that this is actually "The maximum number of items to evaluate (not necessarily the number of matching items)". https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/dynamodb/client/query.html which is interesting especially since this use case is being built to rerun potentially 10ks of failed runs at a single time.

src/cirrus/management/utils/manage.py
```diff
@@ -0,0 +1,103 @@
...
+        return deployment.get_execution_by_payload_id(payload_id)
+    return deployment.get_execution(arn)
+
+
+def query_filters(func):
+    # reverse order because not using decorators to keep command clean
+    """Available inputs to filter stateDB query"""
+    func = optgroup.option(
+        "--collection-workflow",
+        help="The collection to filter on",
```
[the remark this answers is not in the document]
theodorehreuter: Yes its intended as a primary filter and in fact a required for a query since it is the partition key as laid out here https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Query.html so essential.  That field is also added into cirrus infra at the terraform level here
https://github.com/Element84/filmdrop-aws-tf-modules/blob/678ad61b161dd82b0301e438c2867feedb81b32f/modules/cirrus/base/db.tf#L8 so it'll always be present.  Its not intended as a prefix filter.  Right now after talking with Ian I think we want to only be able to filter on a single value of this field for the time being.

I'm leaving all the transformation of the input parameters into a dynamodb query up to already existing implementations for querying the statedb.  

You are correct that it is a constructed field https://github.com/cirrus-geo/cirrus-geo/blob/ad79e0e2b163ec3f27e2d91f9160ebe5aad9fe75/src/cirrus/lib/statedb.py#L712.  I'm not sure the logic behind constructing these in this way from the `payload_id`.

src/cirrus/management/utils/manage.py
```diff
@@ -0,0 +1,103 @@
...
+    if payload_id:
+        return deployment.get_execution_by_payload_id(payload_id)
+    return deployment.get_execution(arn)
+
+
+def query_filters(func):
+    # reverse order because not using decorators to keep command clean
+    """Available inputs to filter stateDB query"""
+    func = optgroup.option(
+        "--collection-workflow",
```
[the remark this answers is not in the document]
theodorehreuter: Good catch. You are right.   With payload ids generated here https://github.com/cirrus-geo/cirrus-geo/blob/ad79e0e2b163ec3f27e2d91f9160ebe5aad9fe75/src/cirrus/lib/process_payload.py#L138 and recalling that `collection` comes from an item, I figure you could have multiple items with different collections on different items that are prepended.

src/cirrus/management/commands/manage.py
```diff
@@ -343,3 +292,48 @@ def list_lambdas(ctx, deployment: Deployment, session: Session):
...
+@manage.command("get-records")
+@query_filters
+@pass_deployment
+def get_records(
+    deployment: Deployment,
+    collection_workflow: str,
+    state: str | None,
+    since: str | None,
+    error_prefix: str | None,
+    limit: int = 100,
```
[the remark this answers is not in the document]
theodorehreuter: So this sent me down a bit of a rabbit hole and I'd like your thoughts:

The only solution seems to be
```
    if limit:
        items = statedb.get_items_page(
                collections_workflow=collections_workflow,
                limit=limit,
                **query_args
        )
    else:
        items = statedb.get_items_page(
                collections_workflow=collections_workflow,
                state=state,
                since=since,
                error_begins_with=error_prefix,
    )
   ```

File: src/cirrus/management/utils/manage.py (hunk 7 above)
Comment (theodorehreuter) [reply src/cirrus/management/utils/manage.py:42]: Correct.  Tweaked the wording to be more specific.

File: src/cirrus/management/utils/manage.py [hunk 13]
```diff
@@ -0,0 +1,103 @@
...
+    def wrapper(*args, **kwargs):
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def download_payload(deployment: Deployment, payload_id, output_fileobj):
+    try:
+        deployment.get_payload_by_id(payload_id, output_fileobj)
+    except botocore.exceptions.ClientError as e:
```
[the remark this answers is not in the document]
theodorehreuter: Per our async conversation, removed this function and kept simple `deployment.get_payload_by_id()` in main logic block of `get-records` to facilitate more appropriate error catching.

src/cirrus/management/utils/manage.py
```diff
@@ -0,0 +1,103 @@
...
+        help="Whether or not to load deployment's user vars into environment",
+    )
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+def download_payload(deployment: Deployment, payload_id, output_fileobj):
```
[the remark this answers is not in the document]
theodorehreuter: added types

src/cirrus/management/utils/manage.py
```diff
@@ -0,0 +1,103 @@
...
+    try:
+        deployment.get_payload_by_id(payload_id, output_fileobj)
+    except botocore.exceptions.ClientError as e:
+        # TODO: understand why this is a ClientError even
+        #   when it seems like it should be a NoKeyError
```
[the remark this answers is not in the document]
theodorehreuter: How handling by extract Code from Client error.  If client error is 404 for not found payload, logging to stderr the payload id that was failed to be retrieved.

.pre-commit-config.yaml
```diff
@@ -48,3 +48,8 @@ repos:
         'types_or': [python, pyi]
         args: []
         require_serial: true
+  - repo: https://github.com/igorshubovych/markdownlint-cli
+    rev: v0.42.0
+    hooks:
+    - id: markdownlint
+      exclude: ^(docs/)
```
theodorehreuter, line 55: excluding docs but honestly I think the docs in that folder are old and should be deleted but that feels beyond the scope of the PR

README.md
```diff
@@ -9,7 +9,7 @@ implemented using a scalable architecture deployed on AWS. Cirrus provides the g
 infrastructure for processing, allowing a user to focus on implementing the specific
 processing logic for their data.
 
-![](src/cirrus/docs/images/arch-overview.png)
+![architecture-overview](src/cirrus/docs/images/arch-overview.png)
```
theodorehreuter, line 12: missing alt text was flagged by markdown linter

src/cirrus/management/utils/manage.py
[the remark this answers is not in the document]
pjhartzell: Two comments:

1. A few descriptive words in addition to the payload ID _could_ helpful. Not sure if a typical user will know that bare payload ID output to stderr means that the payload was not found. But the extra text could also get in the way. Your call. Maybe something like:

```
            if e.response["Error"]["Code"] == "404":
                logger.error(
                    "Payload ID '%s' not found in State DB",
                    item["payload_id"],
                )
```

2. I think other `ClientError` errors should at least get logged, rather than silently ignored. Suggest adding something like:


```
            else:
                logger.error(
                    "Error retrieving payload ID '%s': %s",
                    item["payload_id"],
                    e,
                )
```

Thinking more about the above, note that any error other than a `ClientError` will terminate the process. Is that OK? Or do you want to catch all errors and log specific info for those errors where you know the cause, and just generally log that a payload was not retrieved/generated/whatever for all other error cases? I assume a user will want to know all payload IDs that failed here, not just those that failed in a specific way.

src/cirrus/management/commands/manage.py
pjhartzell: Curiosity question: Why is this not also done in the `get-payload` command?

pjhartzell (changes requested):

src/cirrus/management/commands/manage.py
```diff
@@ -343,3 +302,50 @@ def list_lambdas(ctx, deployment: Deployment, session: Session):
...
+    state: str | None,
+    since: str | None,
+    error_prefix: str | None,
+    limit: int | None,
+):
+    """
+    Use query filters to bulk retrieve payloads.
+    Query multiple records from state DB using filter options, retrieve
+    payloads from s3 using payload-id and pipe to stdout as new line json
+    to facilitate stream processig/xargs piping
```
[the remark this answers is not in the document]
theodorehreuter: point taken, its definitely more verbose than the other command doc strings.  Shorted it up with a minor adjustment to your suggested wording.

CHANGELOG.md
```diff
@@ -11,12 +11,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - `CLAIMED` state to StateDB, along with associated `ALREADY_CLAIMED`
   WorkflowEvent. ([#281])
+- `get_records` command to `manage` command group to facilitate bulk
+retrieval of payloads to be piped into other commands for bulk reruns
+of failed jobs ([#305])
```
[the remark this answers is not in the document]
theodorehreuter: linter failure! [shakes fist]

README.md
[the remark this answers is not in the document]
theodorehreuter: good catch.

src/cirrus/management/commands/manage.py
```diff
@@ -343,3 +302,50 @@ def list_lambdas(ctx, deployment: Deployment, session: Session):
...
+        **query_args,
+    )
+    for item in items:
+        try:
+            with BytesIO() as b:
+                deployment.get_payload_by_id(item["payload_id"], b)
+                b.seek(0)
+                payload = json.load(b)
+
+            payload["process"][0]["replace"] = True
```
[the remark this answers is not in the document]
theodorehreuter: Because the `get-payload` command takes a `payload-id` as input while in `get-payloads` we are retrieving records from the state DB to then generate all the payload IDs we want.  You question has made me tweak it to be more explicit.  Now i"m only going to update the env variable we actually  need (`CIRRUS_STATE_DB`) which will hopefully make the code more self documenting.

theodorehreuter:
@jkeifer  - A comment of Preston got me thinking.  Do we actually want to update the `replace` flag in the `get-records` command?  The original ticket called for that but I'm actually leaning towards NOT doing that after thinking about it.  

The command name `get-records` doesn't make it really obvious that you're updating the returned payloads, and as its a CLI command its ready for piping or further manipulation.  Updating the replace field in the process block is trivial with `jq` if you were piping commands together, and altering the payloads actually feels like unexpected behavior from a command who's unique functionality is the query/query filter/bulk retrieval of payloads.

Thoughts on eliminating the `replace` flag update from the command?

src/cirrus/management/utils/manage.py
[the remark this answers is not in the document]
theodorehreuter: Going with moving the error catch to specifically wrap the `get_payload_by_id` call and leave it at that since if we get a payload back it not an unreasonable assumption that its a valid payload and valid dict and wont cause problems in the json load and process block update.

CHANGELOG.md
```diff
@@ -11,12 +11,16 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - `CLAIMED` state to StateDB, along with associated `ALREADY_CLAIMED`
   WorkflowEvent. ([#281])
+- `get_records` command to `manage` command group to facilitate bulk
```
ircwaves, line 14:
```suggestion
- `get_payloads` command to `manage` command group to facilitate bulk
```

src/cirrus/management/commands/manage.py
```diff
@@ -343,3 +302,57 @@ def list_lambdas(ctx, deployment: Deployment, session: Session):
...
+    since: str | None,
+    error_prefix: str | None,
+    limit: int | None,
+):
+    """
+    Retrieve a filtered set of payloads from S3 via querying the state DB for
+    matching payload IDs, altering paylods to enable rerun.
+    """
+    os.environ["CIRRUS_STATE_DB"] = deployment.environment["CIRRUS_STATE_DB"]
+    statedb = StateDB()
```
ircwaves, line 323: High-level: I feel like the new management commands are doing too much for the `commands` module.  The `commands` module houses thin wrappers around implementations in `cirrus.management.deployment.Deployments`.   This enables usage of the implementation from CLI or from another python module.

src/cirrus/management/commands/manage.py
```diff
@@ -343,3 +302,57 @@ def list_lambdas(ctx, deployment: Deployment, session: Session):
...
+                else:
+                    logger.error(
+                        "Error retrieving payload ID '%s': '%s'",
+                        item["payload_id"],
+                        e,
+                    )
+            b.seek(0)
+            payload = json.load(b)
+
+        payload["process"][0]["replace"] = True
```
ircwaves, line 355: I would recommend moving this to an parameter/option, there might be cases where an operator is relying on cirrus to not reprocess items which did succeed.  The option I'd recommend the option default to False, until we see that you do always want this True.

```suggestion
        payload["process"][0]["replace"] = replace
```

ircwaves (changes requested):
Looks good.  Curious about how this handles at scale, but we can certainly take this for a spin, with the changes suggested.

src/cirrus/management/commands/manage.py
theodorehreuter: good point this actually dovetails with the thoughts I was having above.  Having it as a flag makes sense and keeps it from being confusing and that `get_payloads` defaults to just that, getting payloads

src/cirrus/management/commands/manage.py
theodorehreuter: Good point.  Reworked it a good bit to make the command a much lighter wrapper and moved business logic to deployment class.

src/cirrus/management/deployment.py
```diff
@@ -306,3 +310,45 @@ def template_payload(
...
+
+    def yield_payloads(
+        self,
+        collections_workflow: str,
+        limit: int | None,
+        query_args: dict[str, str | None],
+        rerun: bool,
+    ) -> Iterator[dict]:
+        os.environ["CIRRUS_STATE_DB"] = self.environment["CIRRUS_STATE_DB"]
+        statedb = StateDB()
```
ircwaves, line 322: This is works, but `StateDB` can take the `table_name` directly, rather than round tripping through `os.environ`:
```suggestion
        statedb = StateDB(table_name=self.environment["CIRRUS_STATE_DB"])
```

src/cirrus/management/commands/manage.py
```diff
@@ -152,21 +110,21 @@ def run_workflow(
 def get_payload(deployment: Deployment, payload_id: str, raw: bool = False):
     """Get a payload from S3 using its ID"""
 
-    def download(output_fileobj):
+    def download_payload(
+        deployment: Deployment,
+        payload_id: str,
+        output_fileobj: BinaryIO,
+    ):
```
ircwaves, line 117: commented elsewhere (but lost it), but could/should this also leverage the `yield_payload` implementation?

ircwaves (approved):

src/cirrus/management/deployment.py
theodorehreuter: added

src/cirrus/management/commands/manage.py
theodorehreuter: the `yield_payloads` implementation necessitates hitting the stateDB for the query and then sends payloads to get from s3.  Though you are correct that the old implementation at `get_payload` should use the new `fetch_payload` implementation.  Switched it over, which is DRY-ing things up a bit and also makes `get_payload` lighter since it doesn't have its own internal function its calling.  And it uses the more robust error catching too.

pjhartzell (approved):

Issue #309, opened 2025-04-17
Use Fail step error and cause attributes to get error information into update-state


jkeifer:
Currently update-state uses `GetExecutionHistory` to pull the step function's execution to find the error and cause information, because it does not come through in the failure event from eventbridge. This strategy has problems (see #190).

It seems that we could instead use the ability of the `Fail` step type to pass error an cause information through to extract that error information from the payload and pass it along into `update-state` via the eventbridge event.

I would lobby for making this change on main in a backwards-incompatible way, such that implementations would have to update their workflows to propgate that error information and that `update-state` will no longer call `GetExecutionHistory`, instead having an unknown error if the error information is absent from the event.

jkeifer:
If we complete this issue then #190 can be closed.

theodorehreuter:
closed with merge of #311

PR #308, opened 2025-04-17
v1.0.0 Release prep

theodorehreuter:
release prep for 1.0.0

PR #306, opened 2025-04-10
Docs: Updating Lambda docs

theodorehreuter:
Updates docs to reflect the current state of cirrus lambdas

Goals:

1. Rip out now obsolete docs regarding serverless setup for lambdas (not all serverless refs)
2. Rip out obsolete feeder references where feeder means lambda.  Kept reference to the idea of feeders as an input mechanism
3. Add specific explanations for each of the lambdas.
4. CLI documentation will now be available for the sphinx-build to pick it up and display on the final hosted docs.
5. Updated architecture diagram courtesy of @jkeifer

PR #311, opened 2025-04-18
Update how update-state lambda retrieves errors from step function

theodorehreuter:
closing #309 

If the `Fail` state is correctly passing an error out of the step function in the event of a failure and extract the error information from the payload.

Eliminating the calls to go through step function execution history will help prevent throttling issues in the event of large numbers of failed workflows.  

Documentation of change:
- #306 has some additional documentation in the `workflows` readme to explain how to configure a FAIL state to pass errors into the output payload.  
- #308 flags this as a breaking change
- Log message in PR now reminds users to ensure their `state-machine.json` files are properly designed to pass the error out, as the presence of the error in the output payload is dependent on users configuring their state machine definitions and not determined by the default cirrus terraform.

### DIFF
--- a/src/cirrus/lambda_functions/update_state.py
+++ b/src/cirrus/lambda_functions/update_state.py
@@ -9,18 +9,11 @@ from cirrus.lib.enums import SfnStatus
 from cirrus.lib.events import WorkflowEventManager
 from cirrus.lib.logging import get_task_logger
 from cirrus.lib.process_payload import ProcessPayload
-from cirrus.lib.utils import SNSPublisher, SQSPublisher, cold_start, get_client
+from cirrus.lib.utils import SNSPublisher, SQSPublisher, cold_start
 
 cold_start()
 
 logger = get_task_logger("function.update-state", payload=())
-
-# boto3 clients
-SFN_CLIENT = get_client("stepfunctions")
-
-# how many execution events to request/check
-# for an error cause in a FAILED state
-MAX_EXECUTION_EVENTS = 10
 
 INVALID_EXCEPTIONS = (
     "cirrus.lib.errors.InvalidInput",
@@ -176,23 +169,16 @@ def workflow_failed(
 
 
 def get_execution_error(event: dict) -> dict[str, str]:
-    error = None
-
-    # depends on user correctly configuring error handling
-    # in state-machine.json definiton to pass error out
-    try:
-        error = event["detail"]["error"]
-    except KeyError:
-        logger.exception(
-            "Failed to get error.  Check state-machine.json definition to ensure error is correctly passed out of FAIL state",  # noqa: E501
-        )
-
+    error = event["detail"].get("error")
     if error:
-        logger.debug("Error found: '%s'", error)
+        logger.debug("Error found: '%s'", json.dumps(error))
+        error = mk_error(error["Error"], error["Cause"])
     else:
+        msg = "no error was found in event.  Check that 'state-machine.json' schema is configured to pass out errors from FAIL state"  # noqa: E501
+        logger.exception(msg)
         error = mk_error(
             "Unknown",
-            "update-state failed to find a specific error condition.",
+            msg,
         )
     return error
 
--- a/src/cirrus/lambda_functions/update_state.py
+++ b/src/cirrus/lambda_functions/update_state.py
@@ -66,7 +66,7 @@ class Execution:
             if status == SfnStatus.SUCCEEDED:
                 pass
             elif status == SfnStatus.FAILED:
-                error = get_execution_error(arn)
+                error = get_execution_error(event)
             elif status == SfnStatus.ABORTED:
                 pass
             elif status == SfnStatus.TIMED_OUT:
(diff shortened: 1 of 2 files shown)


theodorehreuter:
@jkeifer  I know we were talking about this being a breaking change, but is it really BREAKING?  

if a user doesn't have the error handling correct to output from the step function, we're not raising an error, but sticking with the previously existing pattern of returning an unknown error.  So in that sense this won't actually break any pipelines since a failed run will still just get logged to the stateDB as expected, with a slightly different message in the logs about recommended error handling in the step function. 

 Less of an implementation question the wider handling of an unknown error is remaining unchanged, and more a question about how we describe the change in the changelog.

src/cirrus/lambda_functions/update_state.py
```diff
@@ -175,43 +168,17 @@ def workflow_failed(
...
-        logger.debug("Error found: '%s'", error)
+        logger.debug("Error found: '%s'", json.dumps(error))
+        error = mk_error(error["Error"], error["Cause"])
     else:
+        msg = "no error was found in event.  Check that 'state-machine.json' schema is configured to pass out errors from FAIL state"  # noqa: E501
+        logger.exception(msg)
         error = mk_error(
             "Unknown",
-            "update-state failed to find a specific error condition.",
+            msg,
```
theodorehreuter, line 181: not sure if replicating the nudge here in the message is superior to how it was before with a general unknown error message.  I'd argue this is better since we want to push people to set up their state machines appropriately this will have the message show up in stateDB too which feels like a reasonable push too on top of just having it in logs.

theodorehreuter:
Is this meeting our goal of 'backwards-incompatibility'?  Im not sure it is since a lack of error content will result in the same unknown error as before which won't break any pipelines.  But I'd think that we wouldn't want to actually crash out the pipeline and get statedb records stuck in a `Processing` state for failed executions, and the `update-state` lambda unable to actually update state on failed runs.  Hope is that by having the message in both the logs and state db will be a sufficient nudge.

src/cirrus/lambda_functions/update_state.py
```diff
@@ -175,43 +168,17 @@ def workflow_failed(
...
-        else:
-            logger.warning(
-                "Could not find execution error in last %s events",
-                MAX_EXECUTION_EVENTS,
-            )
-    except Exception:
-        logger.exception("Failed to get stepfunction execution history")
-
+def get_execution_error(event: dict) -> dict[str, str]:
+    error = event["detail"].get("error")
```
theodorehreuter, line 172: went with get to return `None` since looking at existing runs, there always is an error key.  Looking at failed runs there is an `error` key that is just null on succssful runs and null also on failed runs since we're not passing it out.  Preset to None was unnecessary and `dict .get()` will get the `None` set.

src/cirrus/lambda_functions/update_state.py
```diff
@@ -175,43 +168,17 @@ def workflow_failed(
...
     if error:
-        logger.debug("Error found: '%s'", error)
+        logger.debug("Error found: '%s'", json.dumps(error))
+        error = mk_error(error["Error"], error["Cause"])
     else:
+        msg = "no error was found in event.  Check that 'state-machine.json' schema is configured to pass out errors from FAIL state"  # noqa: E501
+        logger.exception(msg)
```
jkeifer, line 178: If you don't take my suggestion below, note that `logger.exception` can only be used within the `except` block of a `try`/`except`. I'm guessing that means this condition is untested?

jkeifer: Of course now I scroll down and to the tests and see it is tested... Regardless of whether it will raise an exception when hit or not, I am pretty certain this line of code is not doing what is expected.

src/cirrus/lambda_functions/update_state.py
```diff
@@ -9,19 +9,12 @@
...
 from cirrus.lib.process_payload import ProcessPayload
-from cirrus.lib.utils import SNSPublisher, SQSPublisher, cold_start, get_client
+from cirrus.lib.utils import SNSPublisher, SQSPublisher, cold_start
 
 cold_start()
 
 logger = get_task_logger("function.update-state", payload=())
 
-# boto3 clients
-SFN_CLIENT = get_client("stepfunctions")
```
jkeifer, line 19: We'll want to make a ticket on the terraform side to remove the IAM policy grant for access to step functions actions.

src/cirrus/lambda_functions/update_state.py
```diff
@@ -175,43 +168,17 @@ def workflow_failed(
...
+        error = mk_error(error["Error"], error["Cause"])
     else:
+        msg = "no error was found in event.  Check that 'state-machine.json' schema is configured to pass out errors from FAIL state"  # noqa: E501
+        logger.exception(msg)
         error = mk_error(
             "Unknown",
-            "update-state failed to find a specific error condition.",
+            msg,
         )
     return error
```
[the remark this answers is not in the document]
theodorehreuter: Like it.

src/cirrus/lambda_functions/update_state.py
theodorehreuter: Gotcha yeah looked into the docs for logging.  Irrelevant with the proposed solution.

src/cirrus/lambda_functions/update_state.py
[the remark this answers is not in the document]
theodorehreuter: This doesn't quite work since there is always an `error` key, its just `null` if there's no error so the `get` will never get to the `DEFAULT_ERROR` since we'll always find the error key.   First inclination was 

```
error = event['detail'].get('error')
if error:
    return error
return DEFAULT_ERROR
```

Asked AI for a simplification and got the `or` suggestion.  Curious for your thoughts on it since I haven't used `or` much this way before but it fits and keeps things simple.   This also feels slightly odd since having it populate the error key everytime and just give it a null value feels slightly odd and made me wonder if there's a way to avoid that.  My first inclination is fairly concise but I am starting to try and avoid direct `None` checks.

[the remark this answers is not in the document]
jkeifer: Yes, of course, my bad. `or` is the way to go here. 👍

> I do wonder if there's a way to avoid the error key being present in the event bridge event if there is no error passed out. But I realize that set up is likely dependent on how user set up and this solution cover either possibility

I'm going to guess the consistent shape of the messages is quite intentional, and I am pretty much certain that one cannot configure eventbridge to produce custom messages from AWS services. Those message formats are all very strictly defined somewhere (I'm sure there's a python lib with pydantic models or the equivalent for all the different message types).

PR #312, opened 2025-04-24
Docs: update workflow docs

theodorehreuter:


docs/src/cirrus/components/workflows/batch.rst
```diff
@@ -21,7 +22,7 @@ Similarly, Batch does not integrate returned payloads with step functions nearly
 as well as Lambda, so Cirrus has a built-in ``post-batch`` task to help with,
 well, post-batch operations. Specifically, ``post-batch`` can pull the Batch
 payload from S3 and return that in the case of a successful Batch execution. In
-the case of a Batch error, ``post-batch`` will scrape the execution logs for an
+the case of a Batch error, ``post-batch`` will scrape the batch logs for an
```
jkeifer, line 25: I perceive this to be an inaccurate change. It scrapes the logs from the execution, the container logs. It does not specifically handle batch errors, and saying "batch logs" makes me think it is looking at something batch-specific, which is really not the case.

docs/src/cirrus/components/workflows/definitions.rst
```diff
@@ -2,47 +2,57 @@ Writing workflow definitions
 ============================
 
 Cirrus workflows, being implemented via `AWS Step Functions`_, are written in
-the `AWS States Language`_. Like all Cirrus components, workflows require both a
-``definition.yml`` file and a ``README.md`` file. Cirrus uses the Serverless
-plugin `serverless-step-functions`_ to underlay the workflow definitions and
-therefore the ``definition.yml`` format is more or less as documented by the
-plugin.
+the `AWS States Language`_. Workflows are defined in ``state-machine.json`` files and require a companion ``README.md``
```
jkeifer, line 5:
```suggestion
the `AWS States Language`_.
```

I think this is all we can say? What a user calls this file is arbitrary, and the requirement for a README is a holdover from the old cirrus.

docs/src/components/workflows/index.rst
jkeifer, line 1: We don't actually need this file.

This other tree with components was used by the old cirrus docs plugin as a place to put component README files so they could be built into the docs. What this allows was 1) for us to document the built in components with README files that could both be grouped with the components and then dumped out via the CLI, and 2) for projects to build their own docs bundle that included documentation from the READMEs for their own custom components. But we don't have this anymore, so this whole extra section could conceivably go away.

I would suggest merging this content with `docs/src/cirrus/components/workflows/index.rst` and removing this file.

theodorehreuter: Gotcha.  Merged the contents of with `docs/src/cirrus/components/workflows/index.rst` and deleted this guy  There was already heavy overlap between the two.  The other files here under functions and tasks can be deleted when those docs are updated.

docs/src/cirrus/components/workflows/definitions.rst
theodorehreuter: applied

PR #315, opened 2025-06-12
fix sns publishing from update_state

ircwaves:
- **remove(lint): markdowncli causes more problems than it fixes**
- **fix(lib): use context manager instead of SNSPublisher.send directly**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- `pymarkdownlnt` via pre-commit hook
+
+### Fixed
+
+- `SNSPublisher.send` being used directly with more than 10 messages in a call was
+  failing at the SNS API.  As direct usage of `send` was not intended, that method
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Removed direct usage of `SNSPublisher.send`, which was not intended, and
+  moved that method to be `SNSPublisher._send` to indicate that
+  intention. ([#315])
```
jkeifer, line 14: Nit: I might suggest adding a description of the bug here, in addition to this perfectly reasonable explanation of the fix for the bug.

jkeifer:
Feel free to also roll any changes required for 1.0.1 prep into this PR so you can cut that release when this is merged.

CHANGELOG.md
ircwaves:
```suggestion
- `SNSPublisher.send` being used directly with more than 10 messages in a call was 
  failing at the SNS API.  As direct usage of `send` was not intended, that method
  has been moved to `_send`, for alignment with its purpose as an internal message
  sending function. ([#315])
```

PR #318, opened 2025-08-25
remove extraneous execution arn parameter from statedb set_processing

ircwaves:
- **fix(statedb): remove unused "exec" attribute from set_processing**

### DIFF
--- a/src/cirrus/lib/events.py
+++ b/src/cirrus/lib/events.py
@@ -185,7 +185,7 @@ class WorkflowEventManager:
     ):
         if isotimestamp is None:
             isotimestamp = self.isotimestamp_now()
-        self.statedb.set_processing(payload_id, execution_arn, isotimestamp)
+        self.statedb.set_processing(payload_id, isotimestamp)
         self._write_timeseries_record(
             payload_id,
             state=StateEnum.PROCESSING,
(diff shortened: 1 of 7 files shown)


src/cirrus/lib/events.py
```diff
@@ -185,7 +185,7 @@ def started_processing(
     ):
         if isotimestamp is None:
             isotimestamp = self.isotimestamp_now()
-        self.statedb.set_processing(payload_id, execution_arn, isotimestamp)
+        self.statedb.set_processing(payload_id, isotimestamp)
```
jkeifer, line 188: This is because CLAIMED is setting the execution arn and we don't need to any longer for this state update, right? If so feel free to resolve, I'm just checking my understanding.

tests/lib/test_statedb.py
```diff
@@ -32,13 +32,10 @@
 STATES = list(StateEnum._member_names_)
 
 
-def create_items_bulk(item_count, fns, **kwargs):
-    if callable(fns):
-        fns = [fns]
+def create_items_bulk(item_count, fn, **kwargs):
```
jkeifer, line 35: Seems like this is confusingly named? Is this more `state_update_bulk`? Whatever, it was here before so don't feel like you have to change it. Just commenting because the name confused me for a moment.

jkeifer (approved):

PR #322, opened 2025-08-27
Add CLI ability to get previous execution, and input or output

ircwaves:


ircwaves:
Abiguity induced by conflation of stepfunctions executions with payload status means that adding this feature would still leave the water muddy.

Issue #323, opened 2025-09-03
Timestream deprecation and new event solution


jkeifer:
See the related https://github.com/Element84/filmdrop-aws-tf-modules/issues/163.

AWS Timestream for Live Analytics has been deprecated, and is not available in new accounts. We need to find a new solution for events that will accommodate the vastly different scales we have across cirrus deployments, without any major operational burden, and that can be efficiently queried as needed to provide metrics and summary statistics to API users and the cirrus dashboard.

One top contender we feel worth investigating first is Cloudwatch Log Insights. With this solution we can write log messages for each event into our log streams (for process and update state). We can include both unique and high-cardinality attribute information in the message, along with aggregation dimensions used to create the metrics. Because this isn't another service, there isn't anything we need to manage or enable to be able to leverage this solution.

Out-of-the-box, querying such metric data is an async process, but documentation seems to indicate that you can setup metrics that you can then query via the normal Cloudwatch metric API.

We need to first test that this is in fact a viable solution. We want to be able to load test Cloudwatch with a large number of metrics, to see how it handles aggregating millions of events and what the query performance is like.

If this works, then we need to create a new eventing class, with event parity with the existing `EventDB` class. It should be behind a env var feature flag, so we can enable/disable it at will via configuration. Once we have the new eventing class in place, we'll need to hook it in everywhere events could be generated.

If this doesn't work, then we need to consider alternatives. Worst case we can fall back to using the AWS Timestream for InfluxDB service.

PR #324, opened 2025-09-03
Add API methods to deployment and CLI

### LINKED ISSUE #321: Support Cirrus API endpoint functionality in the Cirrus CLI
The API currently exposes workflow data and stats not available from the CLI. The CLI should be able to expose any information exposed by the API, but the functionality to produce the workflow data and stats is not decoupled from the API endpoint logic.

pjhartzell:
**Related Issues:**
- Closes #321 

**Summary:**
- Adds the functionality exposed in the cirrus api lambda to methods on the `Deployment` class.
- Exposes these new Deployment class methods in new sub-commands under the `cirrus manage` command. The new sub-commands are:
  - get-workflow-item     Get individual item for a collections/workflow...
  - get-workflow-items    Get items for a collections/workflow from DynamoDB
  - get-workflow-stats    Get aggregate workflow state transition stats...
  - get-workflow-summary  Get item counts by state for a...
  - list-workflows        List available collections/workflows combinations

**Note:** This PR requires a new AWS Systems Manager parameter to be added: `CIRRUS_DATA_BUCKET`. See https://github.com/Element84/filmdrop-aws-tf-modules/pull/167


src/cirrus/management/commands/manage.py
```diff
@@ -314,3 +314,126 @@ def get_payloads(
...
+@click.option(
+    "--since",
+    default=None,
+    help=(
+        "Only include items updated since this relative duration (e.g., 7d, 36h, 15m)"
+    ),
+)
+@click.option(
+    "--limit",
+    default=100000,
```
ircwaves, line 339: Perhaps the default in the deployments class can remain `100000`, but in CLI land, it can be ratcheted down to something that doesn't flood the terminal, say 10 or 100?

src/cirrus/management/commands/manage.py
```diff
@@ -314,3 +314,126 @@ def get_payloads(
...
```
ircwaves, line 383: again, I'd go lower....but maybe there's an argument for this high.

src/cirrus/management/deployment.py
```diff
@@ -351,3 +357,161 @@ def fetch_payload(self, payload_id: str, rerun: bool | None = None):
...
+        except EventsDisabledError as e:
+            raise StatsUnavailableError from e
+
+    def get_workflow_items(
+        self,
+        collections: str,
+        workflow_name: str,
+        state: str | None = None,
+        since: str | None = None,
+        limit: int = 100000,
```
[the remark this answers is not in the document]
pjhartzell: I didn't feel strongly about changing this one, since I changed the CLI limit to 10. In the end, I dropped it to 10 to match the CLI. If calling as a lib functioin, you'll probably want to override the default, whatever it is.

src/cirrus/management/deployment.py
```diff
@@ -351,3 +357,161 @@ def fetch_payload(self, payload_id: str, rerun: bool | None = None):
...
+        self,
+    ) -> dict[str, Any] | None:
+        "Get aggregate workflow state transition stats from Timestream"
+        eventdb = EventDB(self.environment["CIRRUS_EVENT_DB_AND_TABLE"])
+        logger.debug("Getting stats")
+        try:
+            return {
+                "state_transitions": {
+                    "daily": _daily(eventdb.query_by_bin_and_duration("1d", "60d")),
+                    "hourly": _hourly(eventdb.query_by_bin_and_duration("1h", "36h")),
```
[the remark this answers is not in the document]
ircwaves: Great point.  I'll vote for `C`: those functions are aggregations that are very specific to the response structure that `EventDB` returns, so they should be moved there, and imported into both the `api` lambda, and the `deployment` module.  

Anyone endeavoring to replace (eh-hem, me, #323) that class with an alternate solution should provide such an aggregation, which is consistent with the associated query methods.

Going way-to-out-of-scope, I feel like any `EventDB`-like implementation should probably hide the aggregation implementation under the hood.  But that's too much for here, I believe.

src/cirrus/management/deployment.py
```diff
@@ -351,3 +356,108 @@ def fetch_payload(self, payload_id: str, rerun: bool | None = None):
                         payload_id,
                         e,
                     )
+
+    def get_workflows(self) -> dict[str, Any]:
```
jkeifer, line 360: We use parameter store as a source for CLI configuration... I'm wondering if porting this literally is really a good idea. Would it be better to have a parameter store prefix for workflows? The names could be the "workflow ID" and the value could be the configuration?

I don't want to hold this PR up just for this, but I am also a bit wary of doing any thing that's going to further couple us to the catalog.json abomination. We should probably discuss. Perhaps there's a middle ground where we omit this function for now?

src/cirrus/management/deployment_pointer.py
```diff
@@ -18,6 +18,7 @@
 
 # core required vars.  Not exclusive.
 REQUIRED_VARS = {
+    "CIRRUS_DATA_BUCKET",
```
jkeifer, line 21: This is another reason to not use catalog.json. 1) We don't have a reliable way to manage catalog.json, so plenty of deployments may not have it. 2) This will require terraform changes to ensure this is set in the parameters, so we might as well do something better.

tests/management/conftest.py
```diff
@@ -5,23 +5,14 @@
...
 from cirrus.lib.process_payload import ProcessPayload, ProcessPayloads
 from cirrus.management.cli import cli
 from cirrus.management.deployment_pointer import DEPLOYMENTS_PREFIX
 from click.testing import CliRunner
 
-
-@pytest.fixture()
-def timestream_write_client():
-    with moto.mock_aws():
-        yield boto3.client("timestream-write", region_name="us-east-1")
```
jkeifer, line 22: We're not using this to mock timestream for testing??? I mean, apparently not? You can close this if the answer is yes, I just felt like this change is not obvious so I wanted to confirm.

src/cirrus/management/commands/manage.py
```diff
@@ -314,3 +314,126 @@ def get_payloads(
...
+@manage.command("get-workflow-summary")
+@click.argument("collections")
+@click.argument("workflow_name")
+@click.option(
+    "--since",
+    default=None,
+    help=(
+        "Only include items updated since this relative duration (e.g., 7d, 36h, 15m)"
+    ),
+)
```
jkeifer, line 336: It seems like it might be good to make a custom type for this. See https://click.palletsprojects.com/en/stable/parameter-types/#how-to-implement-custom-types.

src/cirrus/management/commands/manage.py
```diff
@@ -314,3 +314,126 @@ def get_payloads(
...
+@click.option(
+    "--limit",
+    default=10000,
+    show_default=True,
+    type=int,
```
jkeifer, line 341: Looking for the above link I found that click has its own int type, `click.INT`. I'm not sure what the difference is in using that vs `int`, but it must have it for a reason: https://click.palletsprojects.com/en/stable/parameter-types/#built-in-types-listing.

src/cirrus/management/commands/manage.py
```diff
@@ -314,3 +314,126 @@ def get_payloads(
...
+        sort_ascending,
+        sort_index,
+    )
+    click.echo(json.dumps(items_page, indent=2))
+
+
+@manage.command("get-workflow-item")
+@click.argument("collections")
+@click.argument("workflow_name")
+@click.argument("itemid")
```
jkeifer, line 425: Is this `itemids`?

src/cirrus/lib/statedb.py
```diff
@@ -25,6 +25,12 @@ def get_state_db():
     return os.getenv("CIRRUS_STATE_DB")
 
 
+def to_current(item: dict[str, Any]) -> dict[str, Any]:
+    item["catid"] = item["payload_id"]
```
ircwaves, line 29: I'm not sure how to best describe this function, but my thought is something like below, IIUC.
```suggestion
    """Compatiblity function for cirrus-dashboard""""
    item["catid"] = item["payload_id"]
```

ircwaves (approved):

src/cirrus/management/deployment.py
ircwaves: Seconded.

pjhartzell: Convinced. Removed as per middle ground.

src/cirrus/lib/statedb.py
pjhartzell: I like. Better than naught. Inserted.

src/cirrus/management/commands/manage.py
pjhartzell: Yeah. Done.

src/cirrus/management/commands/manage.py
pjhartzell: Vetted all the types and made updates where appropriate.

jkeifer (approved):
🚀

PR #327, opened 2025-09-06
Documentation

pjhartzell:


docs/src/cirrus/10_intro.rst
```diff
@@ -10,9 +10,10 @@ both a cost-efficient and performance-efficient manner, regardless if
 processing takes seconds, hours, or longer.
 
 Cirrus consists of two primary pieces:
-- ``cirrus``, a CLI-based project management and deploy tool, and
-- ``cirrus-lib``, a Python library providing a number of useful
-abstractions solving common needs for users writing their own Cirrus components.
+
+- ``cirrus`` -  A CLI-based project management tool.
+- ``cirrus-lib`` - A Python library providing a number of useful
+  abstractions solving common needs for users writing their own Cirrus components.
```
pjhartzell, line 16: @ircwaves @jkeifer Regarding `cirrus-lib` providing...

>a number of useful abstractions solving common needs for users writing their own Cirrus components

Is this correct anymore? I'm not seeing where pulling in `cirrus.lib` when building a Task, Workflow, or Feeder will be useful. As far as Functions, `cirrus.lib` directly supports the built-in functions in `cirrus.lambda_functions`. Do we have use-cases for users creating lambda functions outside of the built-ins?

docs/src/cirrus/10_intro.rst
```diff
@@ -10,9 +10,10 @@ both a cost-efficient and performance-efficient manner, regardless if
 processing takes seconds, hours, or longer.
 
 Cirrus consists of two primary pieces:
```
pjhartzell, line 12: Do these listed pieces still make sense now that deployment has been removed from `cirrus-geo`?
- `cirrus` is a CLI tool provided by the `cirrus-geo` package, but it no longer does project management or deployment.
- I'm not sure `cirrus-lib` provides useful abstractions outside of its use within the `cirrus-geo` package where it supports the "built-in" lambdas and CLI. See [this comment](https://github.com/cirrus-geo/cirrus-geo/pull/327/files#r2327251587) for more thoughts.

When I think about Cirrus, I think about:
1. How am I going to deploy the infra needed to run Cirrus workflows?
    - Handled by the `cirrus` module in the `filmdrop-aws-tf-modules` repo, which uses the "built-in" lambdas defined in `cirrus-geo`.
2. How do I build and deploy tasks and workflows that I want to run?
    - Again, handled by the `cirrus` module in the `filmdrop-aws-tf-modules` repo.
3.  What tools do I have—outside of the AWS Console—to interact with my Cirrus deployment?
    - This is the Cirrus CLI and, perhaps, the current Cirrus API. Both from the `cirrus-geo` package.

So. Maybe the pieces of Cirrus are:
1. The `cirrus` module in `filmdrop-aws-tf-modules` - A Terraform module that: 1) defines the AWS infra for the core Cirrus machinery, and 2) provides mechanisms for users to define the AWS infra that their custom tasks and workflows require to run on said Cirrus machinery.
2. `cirrus-geo` - A Python package defining the functionality of the core Cirrus machinery.
3. The Cirrus CLI - A CLI for interacting with resources in a Cirrus deployment, which happens to be housed in the same repo and package as `cirrus-geo`.

I'm not happy with the list above, but I think we need a fresh explanation of what Cirrus is. The list is a first attempt at that. Also, I understand that Cirrus is now "Bring your own IaC", but is anything other than the `cirrus` module in `filmdrop-aws-tf-modules` being used to deploy AWS infra for Cirrus > v1.0?

docs/src/cirrus/10_intro.rst
jkeifer: Cirrus.lib could be useful for building feeders, actually. But anything that's in here for building feeders could and probably should be extracted into a utils package (like boto3utils) or is already part of aws-lambda-powertools.

docs/src/cirrus/10_intro.rst
jkeifer: > but is anything other than the cirrus module in `filmdrop-aws-tf-modules` being used to deploy AWS infra for Cirrus > v1.0?

Cirrus-geo doesn't have a preference on how you deploy it. But we do have a problem right now in that `filmdrop-aws-tf-modules` has the authoritative definition of key infrastructure components, like the statedb schema. That is frankly not a great situation -- I think we (I) went too far in the cleanup, we should have retained at least cloudformation templates for the minimum required resources to deploy and run cirrus. In fact I'd like to bring those back in some form at some point (but of course when, right?).

That's not to say that cirrus-geo docs shouldn't point to `filmdrop-aws-tf-modules`, but also the way that project is currently configured it would be hard for someone to use the cirrus module by itself.

In short, I don't know that I'd tie this project too tightly to `filmdrop-aws-tf-modules`. But it is worth a mention.

Otherwise, I like the other two.

PR #330, opened 2025-09-17
Move to uv for dependency management

pjhartzell:
**Summary:**

An attempt to add **stac-task** as a dependency for a new feature resulted in dependency resolution failure, perhaps because the base, cli, dev, and docs dependencies are resolved sequentially rather than simultaneously. Seemed an appropriate motivation to move to `uv` and allow `uv` to resolve all extra (cli) and optional (dev, docs) dependencies in one go. I can confirm that the dependency resolution is successful with these changes in place.

**Proposed Changes:**

1. Moves to `uv` for dependency resolution and installation.
2. Updates several test files to conform to `ruff` formatting.

Changelog is updated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,3 +78,11 @@ To upgrade all dependencies:
 ```commandline
 uv sync --upgrade
 ```
+
+### Installing all dependencies
+
+To install all dependencies (base, dev, docs, and cli):
+
+```commandline
+uv sync --all-groups --all-extras
+```
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -63,7 +63,7 @@ docs = [
     "myst_parser",
     "sphinxcontrib-mermaid",
     "sphinx-click",
-    "sphinx-rtd-theme==2.0.0",
+    "sphinx-rtd-theme",
 ]
 
 [project.urls]
--- a/uv.lock
+++ b/uv.lock
@@ -284,7 +284,7 @@ docs = [
     { name = "myst-parser" },
     { name = "sphinx" },
     { name = "sphinx-click" },
-    { name = "sphinx-rtd-theme", specifier = "==2.0.0" },
+    { name = "sphinx-rtd-theme" },
     { name = "sphinxcontrib-mermaid" },
 ]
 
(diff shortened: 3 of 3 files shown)


.github/workflows/python-test.yml
```diff
@@ -1,34 +1,35 @@
...
+    branches:
+      - main
 
 jobs:
   build:
     strategy:
       matrix:
         python-version:
           - "3.12"
+          - "3.13"
```
ircwaves, line 16: 👍🏽

ircwaves: A second thought on this: in a _someone hit the lotto and is on a long vacation_ situation, it could be a blocker to a patch release.  My ideal would be that py3.13 testing is in a separate job (`python-test-next`) that is not required, but is hugely proactive.

CONTRIBUTING.md
ircwaves, line 1: So many good changes.

.github/workflows/python-test.yml
pjhartzell: I think this gets into whether `cirrus-geo` is expected to be used as a library or is it just an lambda function "app" and a CLI. If not a library, should we look at changing pyproject.toml to target a single Python version rather than `>=`?

pyproject.toml
```diff
@@ -26,10 +26,44 @@ classifiers = [
...
+    "pip-tools",
+    "pre-commit",
+    "pre-commit-hooks",
+    "pymarkdownlnt",
+    "ruff",
+    "types-python-dateutil",
+]
+docs = [
+    "sphinx",
+    "myst_parser",
```
jkeifer, line 63: I don't think we need this anymore. It was originally added to allow markdown README files for project tasks and workflows and whatnot to be consolidated into the docs when building for a specific project. Now that cirrus "projects" are no longer a thing, we don't need to support markdown documentation in sphinx, aside from what markdown-format docs sources we currently have. If we were to cleanup/reformat those to rst we could remove this dep and its configuration.

Note that I'm not suggesting adding more to this PR, I merely wanted to document this point somewhere.

jkeifer (changes requested):

pyproject.toml
```diff
@@ -26,10 +26,44 @@ classifiers = [
...
+    "sphinxcontrib-mermaid",
+    "sphinx-click",
+    "sphinx-rtd-theme==2.0.0",
```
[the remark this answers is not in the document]
pjhartzell: Done.

.github/workflows/docs-build.yml
```diff
@@ -14,27 +14,26 @@ jobs:
...
         run: |
           cd ./docs
-          make clean build CIRRUS_VERSION=$GITHUB_REF_NAME
+          uv run make clean build CIRRUS_VERSION=$GITHUB_REF_NAME
       - name: Copy documentation into gh-pages
         if: github.event_name != 'pull_request'
         run: |
           cd ./docs
-          make gh-pages-copy gh-pages-versions-update CIRRUS_VERSION=$GITHUB_REF_NAME
+          uv run make gh-pages-copy gh-pages-versions-update CIRRUS_VERSION=$GITHUB_REF_NAME
```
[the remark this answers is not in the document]
pjhartzell: Meh. I'm not coming to a strong opinion on this. I think it is good to be explicit with the `uv run make ...`, as it sets the python version and environment that the `make` commands need. More explicit than just activating the environment (`source .venv/bin/activate`). My gut says to keep `uv` out of the makefile, which means keeping it in CI, which means no changes. 

If you have a strong opinion, let me know and I'll make the change.

pyproject.toml
pjhartzell: I'm going to defer on this until I make more progress on getting documentation update. Noted, and will implement then.

.github/workflows/docs-build.yml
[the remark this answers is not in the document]
jkeifer: `uv run` simply being an exec in the venv context seems like it should have been obvious, but for some reason I didn’t realize that was the case. Disregard my concern here, your arguments are solid.

jkeifer (approved):

PR #331, opened 2025-09-19
Use stac-task's Payload model

pjhartzell:
**Background:**
- In the [**stac-task**](https://github.com/stac-utils/stac-task) repo, the "payload" data model component was recently extracted from the `Task` class into its own `Payload` class. 
- A "process payload" in **cirrus-geo** assumes the same structure as the **stac-task** `Payload`, with a few additional restrictions. 
- The functional guts of a Cirrus "task" is [recommended/assumed](https://github.com/Element84/filmdrop-aws-tf-modules/blob/58473ba5c28ebacb13902f003b8c13be6a37c9da/modules/cirrus/README.md#cirrus-task-infrastructure-vs-stac-task-code) to be built with **stac-task**, making a common data model for payloads an implicit requirement.

**Proposed Changes:**

This PR is the realization of making the implicit requirement of a common payload structure between **stac-task** and **cirrus-geo** explicit. It introduces a `CirrusPayload` class that inherits from the **stac-task** `Payload` class as the data model for Cirrus workflow input payloads that are used to instantiate a Cirrus `ProcessPayload`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 
 - Switch to `uv` for package management. ([#330])
+- Use the `Payload` model from `stac-task`. ([#331])
 
 ### Fixed
 
--- a/src/cirrus/management/deployment.py
+++ b/src/cirrus/management/deployment.py
@@ -272,10 +272,10 @@ class Deployment:
             dict containing output payload or error message
 
         """
-        payload = ProcessPayload(payload)
-        wf_id = payload["id"]
+        process_payload = ProcessPayload(payload)
+        wf_id = process_payload.payload["id"]
         logger.info("Submitting %s to %s", wf_id, self.name)
-        resp = self.process_payload(json.dumps(payload))
+        resp = self.process_payload(json.dumps(process_payload.payload))
         logger.debug(resp)
 
         state = "PROCESSING"
@@ -289,7 +289,7 @@ class Deployment:
         execution = self.get_execution_by_payload_id(wf_id)
 
         if state == "COMPLETED":
-            output = dict(ProcessPayload.from_event(json.loads(execution["output"])))
+            output = ProcessPayload.from_event(json.loads(execution["output"])).payload
         elif state == "PROCESSING":
             output = {"last_error": "Unkonwn: cirrus-mgmt polling timeout exceeded"}
         else:
--- a/README.md
+++ b/README.md
@@ -11,26 +11,28 @@ processing logic for their data.
 
 ![architecture-overview](docs/src/cirrus/images/arch-overview.png)
 
-As input, Cirrus takes a STAC ItemCollection, with a process definition block.
-That input is called a Cirrus `ProcessPayload` (CPP).
+As input, Cirrus takes a STAC ItemCollection along with a `process` definition block.
+That input is called a "payload" and follows the `Payload` model defined in the
+[**stac-task**](https://github.com/stac-utils/stac-task) package, with slightly tighter requirements on the presence and content
+of the `process` definition block.
 
-An input is run through a workflow that generates one or more output STAC Items.
+An input payload is run through a workflow that generates one or more output STAC Items.
 These output Items are added to the Cirrus static STAC catalog in S3,
 and are also broadcast via an SNS topic. Subscriptions to that topic can
-triggering additional workflows or external processes, such as indexing into a
+trigger additional workflows or external processes, such as indexing into a
 STAC API catalog (e.g.,
 [stac-server](https://github.com/stac-utils/stac-server)).
 
-Cirrus workflows range from the simple publishing of unmodified input items to
-the complex transformation of input Items and generation of wholly-new output
-Items. The current state of CPP processing is tracked in a state database
-to prevent duplicate processing and allow for a user to follow the state of any
-input through the pipeline.
+Cirrus workflows range from the simple publishing of unmodified input items to the
+complex transformation of input Items and generation of wholly-new output Items. The
+current state of a payload in a processing pipeline is tracked in a state database to
+prevent duplicate processing and allow for a user to follow the state of any input
+payload through the pipeline.
 
 As shown in this high-level overview of Cirrus, users input data to Cirrus
 through the use of _feeders_. Feeders are simply programs that get/generate
 some type of STAC metadata, combine it with processing parameters, and pass it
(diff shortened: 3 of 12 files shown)


pjhartzell:
This PR was initially opened on a forked repo and was reviewed there. See that PR for the review comments: https://github.com/pjhartzell/cirrus-geo/pull/1

src/cirrus/management/deployment.py
```diff
@@ -272,10 +272,10 @@ def run_workflow(
             dict containing output payload or error message
 
         """
-        payload = ProcessPayload(payload)
-        wf_id = payload["id"]
+        payload_manager = PayloadManager(payload)
+        wf_id = payload_manager.payload["id"]
         logger.info("Submitting %s to %s", wf_id, self.name)
-        resp = self.process_payload(json.dumps(payload))
+        resp = self.enqueue_payload(json.dumps(payload_manager.payload))
```
jkeifer, line 278: Do you need `PayloadManager` here, or would `CirrusPayload` suffice?

src/cirrus/management/deployment.py
```diff
@@ -289,7 +289,7 @@ def run_workflow(
         execution = self.get_execution_by_payload_id(wf_id)
 
         if state == "COMPLETED":
-            output = dict(ProcessPayload.from_event(json.loads(execution["output"])))
+            output = PayloadManager.from_event(json.loads(execution["output"])).payload
```
jkeifer, line 292: Same here: do you need `PayloadManager`, or just `CirrusPayload`?

jkeifer: I guess the `from_event` is currently specific to `PayloadManager`. I feel conflicted, because that seems like it could be just as useful on `CirrusPayload`. But if we move everything on to `CirrusPayload` then why did we separate them... 😅

So it looks like the answer to my original question is yes, you do need `PayloadManager`.

jkeifer (approved):
Two minor questions to consider, but I think this looks good enough as-is to approve.

ircwaves (approved):
Approved with suggestions (and 2 comments that I deleted, but maybe will show up :shrug: thanks GH).

docs/src/cirrus/20_arch.rst
```diff
@@ -8,11 +8,11 @@ An overall picture of Cirrus looks like this:
 
 .. image:: images/arch-overview.png
 
-As shown in that diagram, input Cirrus Process Payloads are generated by a :doc:`Feeder <components/feeder/index>` and published into an
-input SQS queue, either manually by a user or from an externals source such as an SNS notification. These enqueued messages are
-then processed by
-the ``process`` Lambda Function. The ``process`` function creates or updates the
-:doc:`State DB <70_statedb>` record for the payload and starts a
+As shown in that diagram, input Cirrus Process Payloads are generated by a :doc:`Feeder
```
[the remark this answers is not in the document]
pjhartzell: Nah. Cleaned up all references to Cirrus Process Payload and CPP.

src/cirrus/lib/payload_manager.py
```diff
@@ -36,69 +37,35 @@ class TerminalError(Exception):
...
+        self.payload = CirrusPayload(
+            *args,
+            **kwargs,
+            set_id_if_missing=set_id_if_missing,
+        )
+        self.payload.validate()
 
+        self.logger = get_task_logger(__name__, payload=self.payload)
+        self.process = self.payload.process_definition
+        self.features = self.payload.items_as_dicts
```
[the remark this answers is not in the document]
pjhartzell: Yep, yep. Already taken care of.

src/cirrus/lib/cirrus_payload.py
```diff
@@ -0,0 +1,65 @@
...
+
+        if "workflow" not in self.process_definition:
+            raise ValueError(
+                "Payload must contain a 'workflow' field specifying the workflow name",
+            )
+
+        if "workflow-" not in self["id"]:
+            raise ValueError(
+                f"Payload 'id' field must contain 'workflow-': {self['id']}",
+            )
```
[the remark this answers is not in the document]
pjhartzell: Might as well correct this now. Done.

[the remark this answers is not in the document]
pjhartzell: Might as well take care of this now. Done.

bin/build-lambda-dist.bash
```diff
@@ -33,8 +34,10 @@ make_zip_base() {
     (
         cd "${site_packages}"
         zip -r "${dest}" .
-        # we don't need or want pip cluttering up our lambda zip
-        zip --delete "${dest}" 'pip/*'
+        # we don't need or want pip cluttering up our lambda zip (if it exists)
+        if zip -sf "${dest}" | grep -q "pip/"; then
+            zip --delete "${dest}" 'pip/*'
+        fi
```
jkeifer, line 40: Do you need a conditional check on this? Does it error if it doesn't delete anything? Would it be better just to swallow the error, you know, ask for forgiveness instead of permission? If so, you could:

```suggestion
        # we don't need or want pip cluttering up our lambda zip (if it exists)
        zip --delete "${dest}" 'pip/*' ||:
```

CHANGELOG.md
```diff
@@ -7,10 +7,18 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## Unreleased
 
+### ⚠️ Breaking changes
+
+- `ProcessPayload` class renamed to `PayloadManager` to better reflect its purpose. The
+  renamed class is also simplified: `features` and `process` attributes removed,
+  `from_event` method moved to the new `CirrusPayload` class. ([#331])
+
```
jkeifer, line 15: I'd push back on this being a breaking change. `ProcessPayload` is an internal model that should not be used outside cirrus-geo. If this were not the case and we did have external users I'd agree this change would be breaking. But as it stands I don't think it is breaking, and would rather see this be listed as a non-breaking change. I might also specify this change applies to an internal class, i.e., "Internal `ProcessPayload` class renamed to `PayloadManager`...".

I'd almost go so far as to argue we don't even need to include this change here, but I generally defer to the side of transparency risking useless info than to potentially hide something meaningful from someone.

jkeifer: If we cannot agree that this should not be classified as a breaking change then consider that we will need to have a major version bump for this release.

jkeifer (changes requested):
Two comments, one super minor and one rather major but not related to the code changes themselves. All in all this is great, thank you for pulling all this together!

jkeifer (approved):

PR #337, opened 2025-09-25
Switch build system to hatchling and fix readme inclusion

jkeifer:
I noticed the README was not rendered as a markdown document on pypi. In fixing that I figured I'd also switch to the more modern build system of `hatchling` instead of `setuptools`.

I also relaxed the version pin on stac-task.


pjhartzell (approved):
Looks good.

PR #338, opened 2025-09-29
CloudFormation

pjhartzell:

PR #342, opened 2025-10-02
Better build script

jkeifer:
PR onto #338 to parametrize the lambda zip build script for runtime python version and arch. Updates gh actions workflow to use a matrix to build for all supported python runtimes and architectures. Includes updates to the cloudformation templates to parametrize the lambda python runtime version and arch correspondingly.

The diff does not reflect the changeset as intended as #338 is from a branch outside this repo and cannot be selected as the PR target. Look at the latest commit for a better diff.


cloudformation/core/functions.yaml
```diff
@@ -291,14 +299,14 @@ Resources:
     Type: AWS::Lambda::Function
     Properties:
       FunctionName: !Sub '${ResourcePrefix}-api'
-      Runtime: python3.12
+      Runtime: !Sub 'python${LambdaPythonVersion}'
       Handler: api.lambda_handler
-      Code: lambda-packages/cirrus-lambda-dist-py3.12-arm64.zip
+      Code: !Sub 'lambda-packages/cirrus-lambda-dist_python-${LambdaPythonVersion}_${LambdaArch}.zip'
```
pjhartzell, line 304: We can't use `!Sub` (or `!Ref`) here if we want to the `package` command to find the local zip file and upload it to S3 for us. We need to either move back to a fixed file name, or manually upload the zip file to S3 and then specify an `S3Bucket` and `S3Key` here.

I'd prefer for `package` to continue to upload our zip.

Issue #355, opened 2025-11-13
`update-state` fails to parse Step Function events when `detail.input` is null or `detail.output` is invalid (truncated)


fabito:
The `update-state` Lambda function is currently failing to process "Step Functions Execution Status Change" events when the event `input` property is null.

Specifically, the `Execution.from_event` method appears to assume `detail.input` is always a populated JSON string or object. When a workflow is triggered with no input, CloudWatch events set this property to null, causing the parser to throw an exception.

Is my current workflow implementation incompatible with the expectations of the `update-state` function ? It is unclear if cirrus-geo strictly requires input data to function  or if the `update-state` function is simply not handling this scenario.

Example payload:

```json
{
    "version": "0",
    "id": "b9756528-24e7-343c-61d6-27539cd7d935",
    "detail-type": "Step Functions Execution Status Change",
    "source": "aws.states",
    "account": "111111111111",
    "time": "2025-11-13T08:59:13Z",
    "region": "ap-southeast-2",
    "resources": [
        "arn:aws:states:ap-southeast-2:1111111111:execution:s1-grd-to-stac:ef4b39d0-bc27-587e-8201-9c408ef78f98"
    ],
    "detail": {
        "executionArn": "arn:aws:states:ap-southeast-2:1111111:execution:s1-grd-to-stac:ef4b39d0-bc27-587e-8201-9c408ef78f98",
        "stateMachineArn": "arn:aws:states:ap-southeast-2:11111111:stateMachine:s1-grd-to-stac",
        "name": "ef4b39d0-bc27-587e-8201-9c408ef78f98",
        "status": "SUCCEEDED",
        "startDate": 1763023536112,
        "stopDate": 1763024353235,
        "input": null,
        "output":  { ... }
    }
}
```

fabito:
That's exactly what I'm doing, perhaps my StepFunction is incorrectly setup ?

This the workflow template:


```json
{
    "Comment": "Example workflow using parallel to make a 'batch group'",
    "StartAt": "batch-group",
    "States": {
        "batch-group": {
            "Type": "Parallel",
            "Branches": [
                {
                    "StartAt": "pre-batch",
                    "States": {
                        "pre-batch": {
                            "Type": "Task",
                            "Resource": "arn:aws:states:::lambda:invoke",
                            "Parameters": {
                                "FunctionName": "${PreBatchFunctionArn}",
                                "Payload.$": "$"
                            },
                            "Next": "batch-task",
[...]

PR #356, opened 2025-11-13
refactor(update-state): improve exception logging during event parsing

fabito:
Currently, when `Execution.from_event` fails, the function catches the exception and re-raises a generic "Unknown event" error. This swallows the original stack trace and error type (e.g., TypeError), making debugging difficult.

This change logs the specific error message in the exception to improve observability.|

See: https://github.com/cirrus-geo/cirrus-geo/issues/355


jkeifer:
Addressed in #358.

PR #357, opened 2025-11-13
feat(post-batch): Make aws batch log group name configurable

fabito:


fabito:
My use case is simple. I was using an existing batch job definition which route logs to an existing log group.
I had to comment out the log configuration, for now, so that logs would become visible to post-batch function.

```hcl
    logConfiguration = {
      logDriver = "awslogs"
      # NOTE: Let it go to /aws/batch/job log group for Cirrus compatibility
      #   options = {
      #     "awslogs-group"         = aws_cloudwatch_log_group.this.name
      #     "awslogs-region"        = data.aws_region.current.name
      #     "awslogs-stream-prefix" = "prefix${each.value.name_suffix}"
      #   }
    }
```

jkeifer:
I don't know that the proposed changes are quite sufficient to fully accommodate this need. [Other places in cirrus have a similar assumption about the batch task log group](https://github.com/cirrus-geo/cirrus-geo/blob/main/src/cirrus/management/task_logs.py#L99) and I'm not sure if there's a trivial and performant way to dynamically retrieve that configuration everywhere it might be required.

I recognize the need here, I just think this needs more investigation to allow a more robust implementation. I've created an issue to track the need: #372.

PR #358, opened 2025-11-14
refactor(update-state): Improve error message when input is not included in event payload

fabito:
Currently when the input is truncated (details.input = null), the `update-state` function fails and the workflow stays in the `PROCESSING` state.

Related to: https://github.com/cirrus-geo/cirrus-geo/issues/355

### DIFF
--- a/src/cirrus/lambda_functions/update_state.py
+++ b/src/cirrus/lambda_functions/update_state.py
@@ -102,7 +102,8 @@ class Execution:
                 error=error,
             )
         except Exception as e:
-            raise Exception(f"Failed to parse event: {e} | Event: {json.dumps(event)}") from e
+            error_msg = f"Failed to parse event: {e} | Event: {json.dumps(event)}"
+            raise Exception(error_msg) from e
 
 
 def workflow_completed(


src/cirrus/lambda_functions/update_state.py
fabito, line 68: FYI, I suspect `output` can also be truncated when the message limits are exceeded

jkeifer: That's a good call out. We depend on tasks ensuring they push their outputs to S3 when they exceed the message length. For batch tasks that's implicit in the task interface, and post-batch will ensure the payload is passed as a url reference. For lambda-based tasks, I'm not sure if we have a clear way to enforce the limit in stac-task.

I have some in-progress changes to lower the length limit to 120KB, which use the escaped serialized length instead of the non-escaped length. But that only affects the input payloads and anything post-batch is handling.

I am coming back to the idea that fetching oversize payloads could be done via the SFN API as you suggested, but I'd want to make that behavior configurable so we can disable it in high-volume pipelines. Of course the code path wouldn't hit if users keep the lengths short anyway, so maybe I rejected the idea too fast.

fabito: I think that hiding it behind a flag (disabled by default) is a good idea.

PR #361, opened 2025-12-16
Add lambda request id to pre/post batch logs

pjhartzell:
**Related Issue(s):**

- None

**Proposed Changes:**

Enables filtering pre- and post-batch lambda logs by the Lambda request ID (aka `aws_request_id`) by injecting the request ID into emitted logs.

1. Adds an `aws_request_id` keyword argument to the `get_task_logger` function and pushes the value into logs if a value other than `None` is passed.
2. Adds a `reset_extra` method to the logging adapter subclass that enables new payload info and lambda request ids to be added to the logger "extra" without having to instantiate a new logger.
3. Grabs the `aws_request_id` attribute from the pre-batch and post-batch lambda handler context object and passes it to the `reset_extra` method (the logger is instantiated at the module level).
4. Adds docs explaining the problem of interleaved lambda logs and how to filter log streams with a lambda request id.

**PR Checklist:**

- [x] I have added my changes to the [CHANGELOG](https://github.com/cirrus-geo/cirrus-geo/blob/main/CHANGELOG.md) **or** a CHANGELOG entry is not required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### Changed
 
-- `DynamicLoggerAdapter` now named `LambdaLoggerAdapter` ([#361])
+- `DynamicLoggerAdapter` now named `CirrusLoggerAdapter` ([#361])
 - `Deployment.get_execution()` now requires an execution ARN; use `get_execution_arn()`
   to resolve an ARN from a payload ID ([#353])
+
+## Removed
+
+- `cirrus.lib.logging.get_task_logger` function; use `CirrusLoggerAdapter` directly
+  ([#361])
 
 ## [v1.2.0] - 2025-11-19
 


src/cirrus/lib/logging.py
```diff
@@ -60,9 +60,16 @@ def process(self, msg, kwargs):
...
     _logger = logging.getLogger(*args, **kwargs)
-    return DynamicLoggerAdapter(_logger, payload, keys=["id", "stac_version"])
+    extra = dict(payload)
+    if aws_request_id is not None:
+        extra["aws_request_id"] = aws_request_id
+    return DynamicLoggerAdapter(
+        _logger,
+        extra,
+        keys=["id", "stac_version", "aws_request_id"],
+    )
```
jkeifer: This seems like a bit of a mess. I think we could simplify this _a lot_ by reading the docs: https://docs.python.org/3/library/logging.html#loggeradapter-objects.

Short version: I think we could make it so 1) payload is an optional dict-like object, 2) the "extra keys" are pulled out of the payload in this function, 3) the extra is passed in as it is but keys are not, and 4) we set `merge_extra=True` on the `LoggerAdapter` in the super init call.

Then we get rid of the process override too. And I don't know we even need the `DynamicLoggerAdapter` class at all; we should be able to do the `self.logger.parent.propagate = False` check in this constructor function.

jkeifer: The reason I looked into this: it might be an instance of premature optimization, but I was wondering if we had to re-instantiate the logger every lambda execution. It seems like all we need to do is reset the extra...

If we are still targeting 3.12 we can't do exactly what I was proposing and we need a `LoggingAdapter` subclass, at least to do the extra merge. But we could extend that subclass to manage our extra too...

...maybe like this "totally written in the comment editor and not well considered" code?

```python
class LambdaLoggerAdapter(logging.LoggerAdapter):
    def process(self, msg, kwargs):
        if self.extra:
            # this does the merge
            kwargs["extra"].update(self.extra)
        return (msg, kwargs)

    def reset_extra(
        self,
        payload: abc.Mapping | None = None,
        aws_request_id: str | None = None,
        **kwargs,
    ) -> None
        self.extra.clear()

        if payload_id := payload.get('id') if payload else None:
            kwargs[id] = payload_id

        if payload_version := payload.get('stac_version') if payload else None:
            kwargs[stac_version] = payload_version

        if aws_request_id:
            kwargs[aws_request_id] = aws_request_id
            
        self.extra.update(kwargs)
```

Then `get_task_logger` could be like:

```python
def get_task_logger(
    *args,
    payload: abc.Mapping | None = None,
    aws_request_id: str | None = None,
    **kwargs,
) -> LambdaLoggerAdapter:
    adapter = LambdaLoggerAdapter(logging.getLogger(*args, **kwargs))
    adapter.reset_extra(payload, aws_request_id)
    return adapter
```

And in functions where we use it we could just use `reset_extra` on every invocation.

Of course, again, this might be premature optimization. I've been burned by instantiating things in handlers before, I think I might be allergic to it now, even where it isn't a problem. Maybe we just make a new adapter for each execution, which means maybe my no-subclass idea is better. Or, if we need 3.12 compat, we subclass just to do the extra merge as I am proposing with the `LambdaLoggerAdapter.process` above.

